PR #4, opened 2015-04-17
Split tests and make them use promises.

Saulis:


Saulis:
I'll be back later with this.

PR #9, opened 2015-05-19
Fixes grid fixed height after sorting, enables redrawer tests.

Saulis:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/9)

### DIFF
--- a/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
+++ b/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
@@ -439,7 +439,15 @@ public class GridComponent implements SelectionHandler<Object>, EventListener,
             jsSort.push(sortOrder);
         }
         $(container).trigger("sort");
-        refresh();
+
+        clearDataSourceCache();
+    }
+
+    private void clearDataSourceCache() {
+        GridDataSource dataSource = getDataSource();
+        if(dataSource != null) {
+            dataSource.clearCache(null);
+        }
     }
 
     public boolean isWorkPending() {
--- a/vaadin-components-gwt/src/main/webapp/vaadin-grid/test/grid-redrawer.html
+++ b/vaadin-components-gwt/src/main/webapp/vaadin-grid/test/grid-redrawer.html
@@ -128,6 +128,23 @@
         });
       });
 
+      describe('grid with a fixed height (issue #8)', function() {
+        before(function() {
+          grid.columns[0].sortable = true;
+          grid.style.height = "500px";
+        });
+
+        it('should keep the correct height after sorting', function() {
+          var firstNonFrozenHeaderCell = qLocal.bind(this, ".v-grid-header .v-grid-cell:not(.frozen)");
+
+          firstNonFrozenHeaderCell().click();
+
+          return grid.then(function() {
+            expect(qaLocal('.v-grid')[2].style.height).to.equal('100%');
+          });
+        });
+      });
+
       describe('using scaled dimensions', function() {
         beforeEach(function() {
           grid.style.position = 'absolute';
--- a/vaadin-components-gwt/src/main/webapp/vaadin-grid/test/grid-redrawer.html
+++ b/vaadin-components-gwt/src/main/webapp/vaadin-grid/test/grid-redrawer.html
@@ -56,7 +56,7 @@
           tdHeight
 
       beforeEach(function() {
-        inner = qLocal("div.v-grid");
+        inner = qaLocal("div.v-grid")[1];
         tbody = qLight('table tbody');
 
         headers = qaLocal('thead tr').length;
@@ -65,11 +65,21 @@
         tfHeight = height(qLocal('tfoot tr td'));
         thHeight = height(qLocal('thead tr th'));
         tdHeight = height(qLocal('tbody tr td'));
-      })
 
-      it('should draw correct dimensions and row heights by default', function() {
-        assertSameDimensions();
-        assertHeightByRows(2);
+        grid.style.height = "";
+        grid.style.width = "";
+        grid.removeAttribute('rows');
+
+        // need to return thenable object for FF sake.
+        return grid;
+      });
+
+      // currently disabled because the grid.then call jams the test on IE.
+      it.skip('should draw correct dimensions and row heights by default', function() {
+        return grid.then(function() {
+          assertSameDimensions();
+          assertHeightByRows(2);
+        });
       });
 
       it('should redraw correctly after modifying body', function() {
@@ -82,11 +92,13 @@
       });
 
       describe('using limited row visibility', function() {
-        beforeEach(function() {
+        before(function() {
           // Increase the number of rows
           tbody.innerHTML += tbody.innerHTML; //4
           tbody.innerHTML += tbody.innerHTML; //8
           tbody.innerHTML += tbody.innerHTML; //16
+
+          return grid;
         });
 
         it('should redraw only visible rows', function() {
@@ -124,23 +136,22 @@
         grid.style.height = '100px';
 
         return grid.then(function() {
-          assertSameDimensions();
+          return grid.then(function() {
+            assertSameDimensions();
+          });
         });
       });
 
       describe('grid with a fixed height (issue #8)', function() {
-        before(function() {
+        it('should keep the correct height after sorting', function() {
           grid.columns[0].sortable = true;
           grid.style.height = "500px";
-        });
 
-        it('should keep the correct height after sorting', function() {
           var firstNonFrozenHeaderCell = qLocal.bind(this, ".v-grid-header .v-grid-cell:not(.frozen)");
-
           firstNonFrozenHeaderCell().click();
 
           return grid.then(function() {
-            expect(qaLocal('.v-grid')[2].style.height).to.equal('100%');
+            expect(qaLocal('.v-grid')[3].style.height).to.equal('100%');
           });
         });
       });
--- a/vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
+++ b/vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
@@ -770,7 +770,6 @@
           }
         }, 1);
 
-        this._bindResizeListener();
       },
 
       _bindResizeListener: function() {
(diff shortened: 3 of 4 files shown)


vaadin-components-gwt/src/main/webapp/vaadin-grid/test/grid-redrawer.html
```diff
@@ -82,11 +92,13 @@
       });
 
       describe('using limited row visibility', function() {
-        beforeEach(function() {
+        before(function() {
```
Saulis: No need to add MORE lines before each test.

vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
```diff
@@ -288,14 +288,18 @@ public IndexBasedSelectionModel getSelectionModel() {
         return (IndexBasedSelectionModel) grid.getSelectionModel();
     }
 
-    public void refresh() {
+    public void forceRefresh() {
```
tomivirkki: This API feels redundant. refresh(boolean) is still used directly when the value is false, otherwise this is used??? Why not just refresh(true)?

manolo: Agree, I would either leave last API or remove parameter to refresh so as finally we have refresh() and forceRefresh()

vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
```diff
@@ -779,19 +779,26 @@
           this.contentDocument.defaultView.addEventListener('resize', function(e) {
             _this._grid.redraw();
           });
-          _this._grid.redraw();
+
+          _this.then(function() {
```
manolo: To me is clearer to use _this only in inner functions.  So use here this preferebly

vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
Saulis: I would've wanted to have the original `refresh()` method and an overload `refresh(boolean)` but unfortunately GWT compile doesn't apparently allow overloads, so I had to rename the original method to `forceRefresh()`. It can be inlined to replace all the references with `refresh(true)` but since it's about matter of taste I didn't want to do that before I hear some opinions from you guys :smile:

manolo:
- How is supposed to review this? using reviewable or just in github ?
- If we merge this here a batch process will override it from gerrit, so you have to manually commit to gerrit.
- I like PR, could we move components from vaadin gerrit to gerrithub, so as we can use both at the same time?

vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
Saulis: It is an inner function.. right? Or how do you define an inner function?

vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
```diff
@@ -439,7 +443,7 @@ public void sort(SortEvent<Object> event) {
             jsSort.push(sortOrder);
         }
         $(container).trigger("sort");
-        refresh();
+        refresh(false);
```
tomivirkki: Sorting never changes the number of items in the datasource and doesn't require any size recalculation so refresh/redraw shouldn't be needed here at all. A call to getDataSource().clearCache(null); instead should be enough.

vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
manolo: Forgot this comment here, there are more than one nested level here, it's more appropriate in the code below.

vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
```diff
@@ -779,19 +779,26 @@
...
+            });
+          });
         });
         this.$.measureobject.data = "about:blank";
       },
 
       attached: function() {
-        this._grid.attached(this, Polymer.dom(this).querySelector("table"), Polymer.dom(this.root));
+        var _this = this;
+        _this._grid.attached(this, Polymer.dom(this).querySelector("table"), Polymer.dom(this.root));
```
manolo: Here is where I meant with _this and this, I like to see _this in nested functions, not in the same level where _this is defined.

Saulis: Ah OK, I can agree with that.

vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
Saulis: Well if that's true, then we can revert all the changes related with `refresh()`. How certain are you about this? :smiley:

Saulis:
> How is supposed to review this? using reviewable or just in github ?
You can try both and see how they feel to use.

 > If we merge this here a batch process will override it from gerrit, so you have to manually commit to gerrit.
Yes, no biggie.

> I like PR, could we move components from vaadin gerrit to gerrithub, so as we can use both at the same time?
Didn't gerrithub close the PRs when you import them?

---

Review status: 0 of 6 files reviewed, 3 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/9)

vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
tomivirkki: Works for me (tried manually and programmatically) and all the tests seem to pass.

Saulis:
Review status: 0 of 6 files reviewed, 3 unresolved discussions, all commit checks successful.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 446 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/9#-JphWYBSYqF5EIFhHSjk-r1-446)** ([raw file](https://github.com/vaadin/components/blob/76e26c9c7712606b13ca736c8613b638852a34a0/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L446)):</sup>
Excellent, I'll modify the changes accordingly. 

I'm going to do a small experiment and amend the changes and force push the new commits just to see how badly it breaks the comments in this pull request :smiley: 

Hold on to your :tophat: 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/9)

Saulis:
Review status: 0 of 6 files reviewed, 4 unresolved discussions, some commit checks broke.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 442 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/9#-JphWYBSYqF5EIFhHSjk-r1-446)** ([raw file](https://github.com/vaadin/components/blob/76e26c9c7712606b13ca736c8613b638852a34a0/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L442)):</sup>
Done.

---

<sup>**[vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html, line 778 \[r2\]](https://reviewable.io:443/reviews/vaadin/components/9#-JpkngBP0tE6zn60QHhf)** ([raw file](https://github.com/vaadin/components/blob/ac3655567e32a971fc6358f31eab11ec9a583883/vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html#L778)):</sup>
As an interesting side-note: changing `this` to `_this` breaks stuff, because `_this.contentDocument` seems to be undefined.  So apparently `this` on line 778 doesn't refer to `<v-grid>` object? Any idea where it refers to?

@tomivirkki ?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/9)

tomivirkki:
Review status: 0 of 6 files reviewed, 4 unresolved discussions, all commit checks successful.

---

<sup>**[vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html, line 778 \[r2\]](https://reviewable.io:443/reviews/vaadin/components/9#-JpkngBP0tE6zn60QHhf)** ([raw file](https://github.com/vaadin/components/blob/ac3655567e32a971fc6358f31eab11ec9a583883/vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html#L778)):</sup>
The listener executes in the context of the measure object so that's what this refers to in this case.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/9)

manolo:
Review status: 0 of 6 files reviewed, 4 unresolved discussions, all commit checks successful.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 291 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/9#-JphWYBSYqF5EIFhHSjk-r1-291)** ([raw file](https://github.com/vaadin/components/blob/76e26c9c7712606b13ca736c8613b638852a34a0/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L291)):</sup>
GWT compiler allows overload, but not in JsInterop classes. So in order to make it work create a new refresh() method and mark refresh(boolean) as @JsNoexport 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/9)

tomivirkki:
Review status: 0 of 6 files reviewed, 2 unresolved discussions, all commit checks successful.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 442 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/9#-JphWYBSYqF5EIFhHSjk-r1-446)** ([raw file](https://github.com/vaadin/components/blob/76e26c9c7712606b13ca736c8613b638852a34a0/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L442)):</sup>
The datasource may be not be defined at this point so a null check is required also

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/9)

Saulis:
Review status: 0 of 4 files reviewed, 2 unresolved discussions, some commit checks pending.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 446 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/9#-JphWYBSYqF5EIFhHSjk-r1-446)** ([raw file](https://github.com/vaadin/components/blob/76e26c9c7712606b13ca736c8613b638852a34a0/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L446)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/9)

PR #36, opened 2015-06-23
Remove unneeded checkbox opacity (fixes #32)

### LINKED ISSUE #32: Wrong checkbox style state
- In the demo, go to example "9" (Selection)
- Select an element then unselect it: the checkbox color weight is higher than others.
- Now scroll a bit down, you will see other checkboxes with the same style.

Basically the reused row (checkbox) keeps the style state which is wrong.

amahdy:
Default color (opacity 0.3) is already specified. 0.6 opacity is different and never seen in the grid because it will be always overriden by the blue color. So practically it's not needed.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/36)

### DIFF
--- a/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/selection/IndexBasedSelectionModelMultiAbstract.java
+++ b/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/selection/IndexBasedSelectionModelMultiAbstract.java
@@ -28,6 +28,7 @@ public abstract class IndexBasedSelectionModelMultiAbstract extends
                         .createCheckInputElement();
                 LabelElement label = Document.get().createLabelElement();
                 checkbox.setId(DOM.createUniqueId());
+                checkbox.setTabIndex(-1);
                 label.setHtmlFor(checkbox.getId());
                 cell.getElement().removeAllChildren();
                 cell.getElement().appendChild(checkbox);
--- a/vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
+++ b/vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html
@@ -310,10 +310,6 @@
     outline: none;
   }
 
-  input[type="checkbox"]:focus + label {
-    border-color: rgba(0,0,0,0.6);
-  }
-
   input[type="checkbox"] + label:after {
     content: url("tick.svg");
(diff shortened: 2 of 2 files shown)


tomivirkki:
+1 for tabindex="-1". This is however a Grid issue and should be reported/fixed there (also).

amahdy:
There is even a slightly bigger issue when you use "TAB" in Vaadin Grid: https://dev.vaadin.com/ticket/18347

amahdy:
-1 for tabindex="-1". It will limit the usability for keyboard gurus.

jounik:
> -1 for tabindex="-1". It will limit the usability for keyboard gurus.

Grid has keyboard navigation and selection by other means (row focus using arrow keys + space for selecting). Having the individual checkboxes focusable reduces the usability.

amahdy:
To add `tabindex=-1`
Possible solutions:

1. Add custom `MultiSelectionRenderer`
  - It requires custom `SelectionModelMulti` and `SelectionMode` as well.
  - `MultiSelectionRenderer` code is different than HEAD in vaadin/grid.

For this solution, `checkBox.setTabIndex(-1)` is needed in `init()` (old code) or `createWidget()` (new code).
2. Fix the issue from vaadin/grid and include a newer widgetset.

manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html, line 338 \[r2\]](https://reviewable.io:443/reviews/vaadin/components/36#-JtCVsul7qxbW6kkjCBh)** ([raw file](https://github.com/vaadin/components/blob/9e88cb4a14d6726d885f1e3afa50c0e70e55eb0f/vaadin-components-gwt/src/main/webapp/vaadin-grid/vaadin-grid.html#L338)):</sup>
Is this needed, how does the user know the checkbox selected when using keyboard?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/36)

manolo:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/36)

Issue #35, opened 2015-06-23
Select all checkbox does not reset


amahdy:
- Create a grid with checkboxes for row selection.
- Click on "select all" top-left checkbox.
- Uncheck rows individually, one by one. (Do not click on the top-left checkbox).

Result: top-left checkbox stays with a semi-selected state.
Expected: should reset to empty selection state.

anezthes:
```Java
    private void updateSelectAllCheckBox() {
        CheckBox selectAllCheckBox = getSelectAllCheckBox();
        if (selectAllCheckBox != null) {
            boolean checked = getSelectionModel().getMode() == IndexBasedSelectionMode.ALL;
            selectAllCheckBox.setValue(checked, false);
            $(selectAllCheckBox).children().addClass("v-grid", "style-scope");
            boolean indeterminate = !(checked ? getSelectionModel().deselected(
                    null, null, null) : getSelectionModel().selected(null,
                    null, null)).isEmpty();
            if (indeterminate) selectAllCheckBox.setValue(false, false);
            // $(selectAllCheckBox).find("input").prop("indeterminate", indeterminate);
        }
    }
```

PR #39, opened 2015-06-25
Issue #35.

anezthes:
This should fix the "all checkbox" state when all and/or none of the grid's items are selected.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/39)

### DIFF
--- a/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
+++ b/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
@@ -52,9 +52,7 @@ import com.vaadin.components.grid.config.JSSortOrder;
 import com.vaadin.components.grid.data.GridDataSource;
 import com.vaadin.components.grid.data.GridDomTableDataSource;
 import com.vaadin.components.grid.data.GridJsFuncDataSource;
-import com.vaadin.components.grid.selection.IndexBasedSelectionMode;
-import com.vaadin.components.grid.selection.IndexBasedSelectionModel;
-import com.vaadin.components.grid.selection.IndexBasedSelectionModelSingle;
+import com.vaadin.components.grid.selection.*;
 import com.vaadin.components.grid.table.GridColumn;
 import com.vaadin.components.grid.table.GridLightDomTable;
 import com.vaadin.components.grid.table.GridStaticSection;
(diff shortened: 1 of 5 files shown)


Saulis:
Just making sure here I understood #35 correctly: the issue happens after user has unselected _all_ rows right?

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 532 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/39#-JsdxiKALCElywKa7qby)** ([raw file](https://github.com/vaadin/components/blob/8b656515b3883177716bd47506594e848d4b2422/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L532)):</sup>
As we discussed, I suggest we encapsulate the logic here to `isIndeterminate` and `isChecked` (for example) methods and place them inside the `IndexBasedSelectionModelMultiAbstract` to simplify this method.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/39)

tomivirkki:
<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 532 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/39#-JsdxiKALCElywKa7qby)** ([raw file](https://github.com/vaadin/components/blob/8b656515b3883177716bd47506594e848d4b2422/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L532)):</sup>
This change results in some strange behavior. If you individually select all the checkboxes, then de-select the "main" checkbox -> selection checkboxes remain selected until you toggle the main checkbox once more. How about instead of having the logic here, add this to indexbasedselectionmodelall:deselect():

            if (size() == 0) {
                clear();
                return false;
            }

And similar check to ....multi:select()

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/39)

Saulis:
There's a test failure (grid-selecting-rows.html, selecting rows by clicking -> should deselect multiple rows) indicating a problem. Tried running the test quickly manually: theres a grid with two rows, and seems that after they've both been selected, they can't be deselected anymore. (the test fires 'mousedown' events to simulate clicking, it might be problem in the test itself if the feature works correctly otherwise)

Otherwise I think the code looks good, after this issue is resolved, we can squash and reword some commits and we're good to go!

---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 55 \[r2\]](https://reviewable.io:443/reviews/vaadin/components/39#-JsvHl0mIF5XJduB0XZv)** ([raw file](https://github.com/vaadin/components/blob/e28781bf14e517d7e7b73d492ed641ecd909b730/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L55)):</sup>
Tip: You might want to tune your IDEA to increase the threshold when replacing imports with *. I have a limit of 10 classes and 5 static classes. Those pesky Eclipse users probably never replace the imports with a *.

Preferences -> Editor -> Code Style -> Java -> Imports -> Class count to use import with '*'

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/39)

PR #41, opened 2015-06-25
Fixed text-overflow: ellipsis for content and header/footer cells

tomivirkki:
Change-Id: I49f88d92c6583693eae743a6ec4aa571b24d3bf8

[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/41)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@
   - "Scrollbar not visible on OS X Safari" (#28)
   - "Chrome OS X hides scrollbars even if System Preferences has 'Show scrollbars always' on" (#30)
   - "Failed to execute write on Document" (#16)
+  - "<v-grid> text-overflow: ellipsis; doesn’t work for cell content" (#10)
 
 ***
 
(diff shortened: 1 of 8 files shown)


jounik:
I would rather not use the same method as in the Framework, since it assumes that all cell content should be inline-blocks.

Did you discuss the option of having a default text renderer which uses `<span>` elements? The header cells could use the same thing: https://trello.com/c/3HSmEC0m/73-default-text-renderer-for-all-cells-should-use-a-span-element-which-handles-text-overflow-ellipsis

Ideally I think we would like to use `display: flex;` for `.v-grid-row`, but then we need to handle the column widths somehow (or is it possible to disable CSS flex altogether?).

anezthes:
We could use flex-grow (for percentage values) and width (for pixel values) to set the column widths.

jounik:
Can we somehow make it backwards compatible with Framework 7 (i.e. browsers without flex box support)?

anezthes:
I guess for those browsers, we either have to use padding and/or line-height for vertical alignment.

Saulis:
Reviewed 8 of 8 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/41)

PR #57, opened 2015-08-04
Fix borders if there are multiple consecutive frozen columns

cwayfinder:
[Screenshot](http://prntscr.com/80n67z)

[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/57)


jounik:
Box shadow avoids the original visual issue. But it would be helpful for designers to have the extra classname to work with, if they need to add some visual treatment which can’t be accomplished with box-shadow.

cwayfinder:
See https://github.com/vaadin/components/pull/62

PR #91, opened 2015-08-18
Enforce max col width (fixes #46)

### LINKED ISSUE #46: <v-grid> Column max-width does not cut content if content is wider than the max-width value
For example, 

given we have a `<col max-width="100">` in use, and the styles `padding-left: 24px` and `padding-right: 28px` applied to the column in question using `td`, the actual Column width calculated by the Grid is something else than `100px`.

There's a fat chance that the Grid assumes that the padding is `24px` on both sides and bases it's calculations on that.

tepi:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/91)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,7 @@
   - Update footer styles (#48)
   - <v-grid> is rendered on top of <paper-drawer-panel> (#79)
   - Prevent infinite size update loop (#33)
+  - Enforce maximum width for expanding columns by cutting content (#46)
 
 ***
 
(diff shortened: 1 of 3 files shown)


tomivirkki:
Missing a test & changes, needs a manual rebase

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/91)

tepi:
Added changes. As for the test, there already was one but a useless one since the content was narrower than the maxWidth set. Fixed it by changing content to be wider.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/91)

tomivirkki:
Reviewed 3 of 3 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/91)

PR #92, opened 2015-08-19
v-grid gets displayed as a flex box (fixes #85)

### LINKED ISSUE #85: v-grid dissapears when used as a flex item
By putting **flex:1** on a v-grid, it will disappear. the ```<v-grid>``` element itself will have the correct size but within it ```<div id="measure">``` and sibling div containing the table will have zero height. The said div will also have the class name v-grid-loading-data which indicates that something went wrong in initialisation.

tomivirkki:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/92)


jounik:
LGTM

PR #93, opened 2015-08-19
Add cross-scope styling support

cwayfinder:
Allow developers to easily modify the theme of v-grid to suit their brand colors etc. By default vaadin-grid should not look like Material Design, that should be an optional addition (or we could provide a "barebone" version of the element).

Use the paper-styles custom properties as the default values for any suitable properties (such as colors): https://github.com/PolymerElements/paper-styles

- [ ] Research the ways how styles can be imported in Polymer (i.e. affecting all component instances, affecting instances with a certain class, affecting only a single instance)
- [ ] Decide the use cases we want to support (e.g. I want to have most grids with the Material Design but some with a custom theme)
- [ ] Design the pieces of the theme that should be customizable (start small)
- [ ] Design and add a new example to the "Sizing and Styling" section which shows one alternate theme

[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/93)

### DIFF
--- a/vaadin-components-gwt/src/main/webapp/vaadin-grid/demo.html
+++ b/vaadin-components-gwt/src/main/webapp/vaadin-grid/demo.html
@@ -4,7 +4,11 @@
 <meta charset="UTF-8">
 <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1" />
 <script src="../bower_components/webcomponentsjs/webcomponents-lite.js"></script>
+<link rel="import" href="ugly-theme.html">
+<link rel="import" href="material-design-theme.html">
 <script src="../VaadinGrid/VaadinGrid.nocache.js"></script>
+
+<!--<style is="custom-style" include="ugly-theme"></style> -->
 
 <script src="sdm.js"></script>
 <script src="demodata.js"></script>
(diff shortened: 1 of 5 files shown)


jounik:
This is an excellent start!

We need to move all the Material Design related styles to the default property/mixin values, and perhaps move that to a separate HTML import people can import if they want that style. By default the grid should inherit styles from the parent page as much as possible (meaning mostly just font styles, and using white as the background color for most things).

jounik:
vaadin-grid.html should not depend on paper-styles.html

To format in another way:

- vaadin-grid.html had rudimentary styles, and defines the mixins and variables which can be used. Only some of the variables need to have a default value, namely the header and footer background colors, since otherwise the body rows will bleed through them.

- We provide another HTML import along vaadin-grid.html, called material-design-theme.html (next to vaadin-grid.html), which defines the v-grid variables and mixins so that it looks like the Material Design data table

- The end user is then responsible of importing the theme along the component, by default the grid will look plain. This means we need to update documentation and examples in many places (haven’t defined those places yet)

jounik:
I believe we need to start smaller after all, and not try to add full-on theming support right away. I will add a new issue for the “start small” approach.

jounik:
https://github.com/vaadin/vaadin-grid/issues/240

PR #112, opened 2015-08-26
Fixes row details problems

tepi:
- Use the proper DetailsGenerator.NULL when undef is given from JS
 - Close all visible details when the generator is changed
 - Fixed setRowDetailsVisible(idx) to actually show the details

[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/112)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 - Added data.getItem api for fetching data items by row index (#45)
 - Combined multi selection models into one model
 - Moved row focus indicator line from the bottom to the left side of a row
+- Added row details feature
 - Issues fixed:
   - Grid doesn't work when using selection-mode multi, frozen columns and sortable columns. (#7)
   - Grid height is recalculated incorrectly when sorting a grid with a fixed height. (#8)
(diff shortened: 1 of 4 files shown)


jounik:
Would be nice if we could avoid these unnecessary white-space changes. Eclipse probably handles those differently than Atom (if someone does edit java files with Atom).

tepi:
I'm using the vaadin framework formatting guidelines for eclipse. Unfortunately it seems that different eclipse versions handle "Remove trailing whitespace on all lines" differently. Mine seems to add that one space to empty block comment lines.

Also, I'm not sure if everyone in the project is using the vaadin conventions at all. We should decide and enforce some conventions to avoid this kind of issues. Be it vaadin formatter or something else as long as everyone uses the same ones.

I'll see if I can modify the patch to remove the white-space changes.

tomivirkki:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/112)

tomivirkki:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/112)

PR #119, opened 2015-09-01
Update row dom order on resize

tepi:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/119)

### DIFF
--- a/vaadin-components-gwt/src/main/java/com/vaadin/client/widgets/Escalator.java
+++ b/vaadin-components-gwt/src/main/java/com/vaadin/client/widgets/Escalator.java
@@ -1301,7 +1301,8 @@ public class Escalator extends Widget implements RequiresResize,
                         cellElem.addClassName("frozen");
                         position.set(cellElem, scroller.lastScrollLeft, 0);
                     }
-                    if (columnConfiguration.frozenColumns > 0 && col == columnConfiguration.frozenColumns - 1) {
+                    if (columnConfiguration.frozenColumns > 0
+                            && col == columnConfiguration.frozenColumns - 1) {
                         cellElem.addClassName("last-frozen");
                     }
                 }
@@ -1568,7 +1569,8 @@ public class Escalator extends Widget implements RequiresResize,
             }
         }
 
-        private void toggleFrozenColumnClass(int column, boolean frozen, String className) {
+        private void toggleFrozenColumnClass(int column, boolean frozen,
+                String className) {
             final NodeList<TableRowElement> childRows = root.getRows();
 
             for (int row = 0; row < childRows.getLength(); row++) {
--- a/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
+++ b/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java
@@ -105,6 +105,8 @@ public class GridComponent implements SelectionHandler<Object>,
         staticSection = new GridStaticSection(this);
 
         grid.setStylePrimaryName("vaadin-grid style-scope vaadin-grid");
+
+        grid.setWidth("100%");
     }
 
     public GridEditor getEditor() {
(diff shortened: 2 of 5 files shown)


tomivirkki:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 433 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/119#-Jy7Q7fQ1gT7mXiDFjC1)** ([raw file](https://github.com/vaadin/components/blob/41b4174422d780f517d8001c718afbb5f13c09be/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L433)):</sup>
All the synchronous methods that called this directly no longer work right (for example grid.columns[0].width = 100). Perhaps they should invoke grid.onResize() instead?

---

<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 434 \[r1\]](https://reviewable.io:443/reviews/vaadin/components/119#-Jy7Ptpby8g7aPJehbyo)** ([raw file](https://github.com/vaadin/components/blob/41b4174422d780f517d8001c718afbb5f13c09be/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L434)):</sup>
I guess this can also be removed now (needs to be called once in the constructor though)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/119)

tomivirkki:
<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 436 \[r2\]](https://reviewable.io:443/reviews/vaadin/components/119#-Jy7vFfDSbwHF4Jl0jJ5)** ([raw file](https://github.com/vaadin/components/blob/ac0ef078000102b76f37fec3beab0aa40483712c/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L436)):</sup>
Well I was kind of thinking about removing this method altogether. We don't need to define a method that only calls another public method.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/119)

tomivirkki:
<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java, line 436 \[r2\]](https://reviewable.io:443/reviews/vaadin/components/119#-Jy7vFfDSbwHF4Jl0jJ5)** ([raw file](https://github.com/vaadin/components/blob/ac0ef078000102b76f37fec3beab0aa40483712c/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/GridComponent.java#L436)):</sup>
Well I was kind of thinking about removing this method altogether. We don't need to define a method that only calls another public method. Additionally now onResize() would be called twice on every updateSize() call.

Would it work if they just called updateSize instead? There's a drawback because it adds a 50ms delay. An alternative would be just to call gridComponent.getGrid().onResize() directly.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/119)

tomivirkki:
lgtm, needs a manual merge/rebase however.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/119)

tomivirkki:
<sup>**[vaadin-components-gwt/src/main/java/com/vaadin/components/grid/ViolatedGrid.java, line 72 \[r4\]](https://reviewable.io:443/reviews/vaadin/components/119#-JyBk1lAoh_V-QNxFSaq)** ([raw file](https://github.com/vaadin/components/blob/ab3f88ff17b5fd3cfb35742c83d2b45721196d96/vaadin-components-gwt/src/main/java/com/vaadin/components/grid/ViolatedGrid.java#L72)):</sup>
The reordering logic is no longer needed for the focus style so this could be reverted. Details row could have a bottom border by itself so the order wouldn't matter there either?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/119)

tomivirkki:
Reviewed 2 of 4 files at r3, 1 of 2 files at r4, 2 of 3 files at r5.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/components/119)

PR #131, opened 2015-09-03
Fix #128 Column hiding button moves horizontally on Firefox when clic…

### LINKED ISSUE #128: Column hiding button moves horizontally on Firefox when clicking it
The column hiding icon moves horizontally a few pixels when the mouse button is pressed down on it.

There’s some weird blurring happening in Safari as well when pressing the button.

cwayfinder:
…king it

[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/131)


jounik:
I think we should use `line-height` and `text-align: center;` (on the `.v-grid-sidebar-button` selector), or flexbox even, to center the icon inside the button. Using absolute positioning will fail when someone wants to change the font size in the header.

But I guess we can change that when we refactor the styles to support cross-scope styling and general theming. All sizes are more or less fixed right now, so these are quite acceptable.

cwayfinder:
BTW, the .vaadin-grid-sidebar-button::before is also positioned with magic numbers.

jounik:
Right. Let’s review those magic numbers when we do the cross-scope styling feature, this should be okay for now.

PR #133, opened 2015-09-03
Fix #125 Column hiding menu is missing styles for keyboard selection

### LINKED ISSUE #125: Column hiding menu is missing styles for keyboard selection
An item in the column dropdown menu is only highlighted on mouse hover. The items are selectable with the arrow keys as well, but the current focus is not shown.

Use the `.gwt-MenuItem-selected` class with the same styles is with hovering.

cwayfinder:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/components/133)


jounik:
Looks good to me.

PR #155, opened 2015-09-14
Fix excessive calls when setting the `frozenColumns` property

cwayfinder:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/155)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ has its own repository.
 - Removed old demos used for developing.
 - Issues fixed:
   - Exception when removing frozen column (#147)
-  - Fix excessive calls when setting the "frozenColumns" property
+  - Fix double setter calls when setting property value
 
 ## Vaadin Grid v0.3.0.beta7 (2015-Sept)
 - Polymer updated to v1.1.1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ has its own repository.
 - Removed old demos used for developing.
 - Issues fixed:
   - Exception when removing frozen column (#147)
+  - Fix excessive calls when setting the "frozenColumns" property
 
 ## Vaadin Grid v0.3.0.beta7 (2015-Sept)
 - Polymer updated to v1.1.1
--- a/java/src/main/java/com/vaadin/components/grid/GridComponent.java
+++ b/java/src/main/java/com/vaadin/components/grid/GridComponent.java
@@ -91,6 +91,8 @@ public class GridComponent implements SelectionHandler<Object>,
 
     public static final int MAX_AUTO_ROWS = 10;
 
+    private int _frozenColumns;
+
     public GridComponent() {
         grid = new ViolatedGrid();
         grid.setSelectionModel(new IndexBasedSelectionModelSingle());
@@ -218,7 +220,11 @@ public class GridComponent implements SelectionHandler<Object>,
     }
 
     public void setFrozenColumns(int frozenColumn) {
-        grid.setFrozenColumnCount(JSValidate.Integer.val(frozenColumn, 0, 0));
+        if (getColumns().size() > 0) {
+            grid.setFrozenColumnCount(JSValidate.Integer.val(frozenColumn, 0, 0));
+        } else {
+            this._frozenColumns = frozenColumn;
+        }
     }
 
     public void scrollToRow(int index, String scrollDestination) {
@@ -326,6 +332,10 @@ public class GridComponent implements SelectionHandler<Object>,
                 @Override
                 public void onChange(List<ChangeRecord> changes) {
                     setColumns(cols);
+                    if (_frozenColumns > 0) {
+                        setFrozenColumns(_frozenColumns);
+                        _frozenColumns = 0;
+                    }
                 }
             });
         }
--- a/java/src/main/java/com/vaadin/components/grid/GridComponent.java
+++ b/java/src/main/java/com/vaadin/components/grid/GridComponent.java
@@ -91,8 +91,6 @@ public class GridComponent implements SelectionHandler<Object>,
 
     public static final int MAX_AUTO_ROWS = 10;
 
-    private int frozenColumns;
-
     public GridComponent() {
         grid = new ViolatedGrid();
         grid.setSelectionModel(new IndexBasedSelectionModelSingle());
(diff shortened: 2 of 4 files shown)


manolo:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[vaadin-grid.html, line 1118 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155#-JzAhE7KZ_lH72xkgW1G)** ([raw file](https://github.com/vaadin/vaadin-grid/blob/ac376ff37e107c85b56a6294218c71b5464b6dc0/vaadin-grid.html#L1118)):</sup>
Why do we need to wait for grid to finish? If it's necessary we might  revisit other setters as well, otherwise I think tests could fail if you defer the execution.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155)

cwayfinder:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[vaadin-grid.html, line 1118 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155#-JzAhE7KZ_lH72xkgW1G)** ([raw file](https://github.com/vaadin/vaadin-grid/blob/ac376ff37e107c85b56a6294218c71b5464b6dc0/vaadin-grid.html#L1118)):</sup>
We only have to wait for the first time in case when we use the "frozen-column" attribute in light DOM. The grid tries to set the frozenColumns property before columns were initialised. Otherwise we get IllegalArgumentException http://prntscr.com/8g73x2

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155)

cwayfinder:
BTW, the `if (newValue != oldValue)` check is also necessary. There would be excessive calls when we remove it http://prntscr.com/8g77cy

manolo:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[vaadin-grid.html, line 1118 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155#-JzAhE7KZ_lH72xkgW1G)** ([raw file](https://github.com/vaadin/vaadin-grid/blob/ac376ff37e107c85b56a6294218c71b5464b6dc0/vaadin-grid.html#L1118)):</sup>
IMO, Grid shouldn't do this check. It should set frozen columns attributes even though columns has not been set yet. Let's ask framework about their opinion.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155)

manolo:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[java/src/main/java/com/vaadin/components/grid/GridComponent.java, line 94 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155#-JzEaHvIX3fp9rKrW3Wr)** ([raw file](https://github.com/vaadin/vaadin-grid/blob/15a561eb267e87126de65ac2037a9957d8f6a149/java/src/main/java/com/vaadin/components/grid/GridComponent.java#L94)):</sup>
Since this code is java there is not need for the underscore in the name

---

<sup>**[java/src/main/java/com/vaadin/components/grid/GridComponent.java, line 335 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155#-JzE_WtVBYE5GATqozhp)** ([raw file](https://github.com/vaadin/vaadin-grid/blob/15a561eb267e87126de65ac2037a9957d8f6a149/java/src/main/java/com/vaadin/components/grid/GridComponent.java#L335)):</sup>
Shouldn't this be placed in the setColumns method instead of inside the observer?. 
I also would call setFrozenColumns always, and prevent the error if frozen columns is greater than columns:

       setFrozenColumns(Math.min(frozenColumns, columns.size());

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155)

cwayfinder:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[java/src/main/java/com/vaadin/components/grid/GridComponent.java, line 94 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155#-JzEaHvIX3fp9rKrW3Wr)** ([raw file](https://github.com/vaadin/vaadin-grid/blob/15a561eb267e87126de65ac2037a9957d8f6a149/java/src/main/java/com/vaadin/components/grid/GridComponent.java#L94)):</sup>
Done.

---

<sup>**[java/src/main/java/com/vaadin/components/grid/GridComponent.java, line 335 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155#-JzE_WtVBYE5GATqozhp)** ([raw file](https://github.com/vaadin/vaadin-grid/blob/15a561eb267e87126de65ac2037a9957d8f6a149/java/src/main/java/com/vaadin/components/grid/GridComponent.java#L335)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155)

tepi:
Add test & a mention to change log

cwayfinder:
Apparently the `value` only returns default value and cannot be used as usual getter. 
```
frozenColumns: {
    ...
    value: function() {        
        return ...;
    },
}
```

tomivirkki:
It makes no sense to just fix this for frozen-columns because it's not a frozen-columns -specific issue. The same problem affects all the properties that are updated in attributeChanged ("frozen-columns, selection-mode, disabled, rows" for now).
Thus the proper place to fix this would be in attributeChanged.

cwayfinder:
Done.

manolo:
LGTM, lets see if others have something to say, otherwise merge it.

---

Reviewed 1 of 1 files at r6, 1 of 1 files at r8, 1 of 1 files at r9.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155)

tomivirkki:
It only handles the "default" case in attributeChanged. "rows" and "frozen-columns" work correctly but "disabled" and "selection-mode" are still called twice

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/155)

PR #158, opened 2015-09-14
Fix/removing frozen columns

cwayfinder:
Look for the fix in the second commit

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/158)


jounik:
I think we discussed that we would not bring in the whole copy of Grid.java or Escalator.java into this project directly, but have a dedicated branch in the Framework repo instead. The Framework team will then provide us with a custom build from that branch that we can use directly in our project.

But of course this PR can be used to test that the fix works. Not sure how the process should go in detail. Maybe you can discuss that with @tepi tomorrow?

tepi:
I'll take it to the FW team and handle it

PR #164, opened 2015-09-16
Pr ajax #148

### LINKED ISSUE #148: Replace randomuser.me API with a more stable data source
The randomuser.me service is occasionally unavailable, and that makes both our examples and user tests flakey.

One option is to store a static version of one result set of the API, and serve that from vaadin.com.

manolo:
closes #148

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/164)


jounik:
Would be great to replace the images also with some other service or some static files on vaadin.com.

tepi:
Also, do we really want this file into the vaadin-grid package? Perhaps some other location on our servers would be better. Then we could just link to that.

manolo:
@jounik  Images is not a big problem, since the example works without them when the service is unavailable. The big problem is when the rest call is not available since  the user thinks that the grid is not working.
@tepi I don't see any reason to have this file in our grid demo, it's a normal practice so as the demo always work without depending on any 3rd party service. Take a look to iron-list demo they have a contacts.json : https://github.com/PolymerElements/iron-list/tree/master/demo/data  

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

tomivirkki:
About serving the results as a static json file. Static data works fine with iron-list because they don't support lazy loading. As you scroll the vaadin-grid it might ask the datasource for indexes 90-110, but with static data it'll always get the same results because the provided parameters don't really do anything. After scrolling the grid it will appear as if you always land on the same page.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

tomivirkki:
Additionally, we have some demos where you can filter the results by gender.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

manolo:
IMO, the demo should not depend on any external service, so for parametrised parameters I would do it in JS. Is that option good for you? or do you prefer to host the service in some vaadin location.  The latest implies to have access to an application with a rest service, it might be a servlet or vaadin app, etc.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

manolo:
As a last thought, we might ask the first time to randomuser api, if the request fails we could do a second try to static.
I don't think maintaining a servlet is a good idea though.
 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

tomivirkki:
Perhaps we could have another JS file like demodatasources.js and define a fake remote there that could be used like fetch("http://api.randomuser.me/?results=14") but it actually parses the static users.json file with the given parameters. It could even return the results with a small delay to make it believable.

This would appear to the user as if we're actually fetching from a remote. If he'll start digging deeper and finds the js file, we could have some comments there stating why it's actually faked.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

tomivirkki:
Or similarly as in jQuery: getJSON("http://api.randomuser.me/?results=14", callback)... Currently there's a generated employees array copy-pasted in many of the demos. That could also be moved into demodatasources.js to clean up the demos.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

tomivirkki:
https://github.com/vaadin/vaadin-grid/blob/master/demo/common.html#L13 could be fixed to "latest" in this patch also?

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

tomivirkki:
This still doesn't handle filtering by gender / grid asking users between specific indexes. Now that the data is in JSON it should be pretty straightforward to filter the results by gender/getting a specified range of users.

Also, the overridden XHR doesn't seem to work with Angular 2 demo

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

tomivirkki:
Reviewed 1 of 2 files at r1, 1 of 3 files at r4, 1 of 2 files at r5.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/164)

PR #165, opened 2015-09-16
Fix  #135 - Hiding the header/footer declarative requires the element…

### LINKED ISSUE #135: Hiding the header/footer declarative requires the element to have at least one cell
It should be possible to hide the header or footer by just specifying an empty `<thead hidden>` or `<tfoot hidden>` element. Currently it has no effect. But if you specify also `<tr><td></td></tr>` inside them, then they get hidden.

cwayfinder:
… to have at least one cell

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/165)


tomivirkki:
Missing vaadin-grid.min.js and test case

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/165)

PR #211, opened 2015-10-12
Flatten Grid JS object API

manolo:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/211)


Saulis:
How is it broken and how does this fix it? Is there a test which I can run to see the problem?


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/211)

Saulis:
This requires some investigation, all tests pass in master, but not when running in SDM.

PR #213, opened 2015-10-19
Rename components to elements

Saulis:
connects to vaadin/components-team-tasks#19

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/213)


demo/columns.html
```diff
@@ -64,7 +64,7 @@
   <section>
     <h3>Adding Columns Dynamically</h3>
     <p>Columns can be added dynamically using the JavaScript API. You can also
-       modify the lightDom of the component, but it is not the recommended
```
tomivirkki: component should be ok here

PR #239, opened 2015-11-08
Behaviors

manolo:
This is necessary in order to parse vaadin-grid js-doc information and generate gwt java code with the gwt-api-generator#r_1.2.1 script.

connects to vaadin/components-team-tasks#34

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/239)

### DIFF
--- a/vaadin-grid-doc.html
+++ b/vaadin-grid-doc.html
@@ -2,14 +2,12 @@
 
 <script>
   Polymer.Vaadin = Polymer.Vaadin || {};
-</script>
 
-<!-- Object for controlling and accessing the columns in the grid. -->
-<script>
-/**
- * @polymerBehavior Column
- */
-Polymer.Vaadin.Column = {
+  /**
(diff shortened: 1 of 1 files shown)


jounik:
Is the generator otherwise able to parse and generate APIs for vaadin-grid?

manolo:
It generates the API, but all parameters marked for instance as '@type Column' will generate methods with Object.

Right now when gwt-api-generator parses vaadin-grid-doc it generates a bunch of web-components for instance ColumnElement.java and ColumnWidget.java instead of an JsInterface for the Column JS object.
If we change web-components by behaviors, the generator generates those interfaces.

The only problem is the our jsapi website-generator script which is prepared to use webcomponents as object documentation, but it's easily modifiable.

Saulis:
Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/239)

PR #254, opened 2015-11-11
Fixes #244

### LINKED ISSUE #244: Row details element is 1px too high
If you inspect the example at https://cdn.vaadin.com/vaadin-elements/latest/vaadin-grid/demo/details.html you can see that the `.userdetailswrapper` element is 190px high, but the `.vaadin-grid-spacer` element wrapping it is 192px high (it has 1px border at the bottom, so it’s 1px too high). This is probably some sort of sub-pixel rounding error.

This makes it difficult to style the detail row because it overlaps the following row elements.

Either the wrapper element sizing should be fixed, or the following rows should be shifted down 1px to accommodate the full detail row.

jounik:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/254)

### DIFF
--- a/vaadin-grid.html
+++ b/vaadin-grid.html
@@ -184,6 +184,7 @@ More examples available at https://cdn.vaadin.com/vaadin-elements/latest/vaadin-
     position: absolute;
     border-bottom: 1px solid #e3e3e3;
     z-index: 1;
+    box-sizing: border-box;
   }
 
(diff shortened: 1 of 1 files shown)


jounik:
Scratch that. If you remove the bottom border from the spacer element, it’s again 1px too high.

PR #268, opened 2015-11-23
When calling selectAll() attributes should be updated (2nd approach)

### LINKED ISSUE #247: grid.selection.selectAll() doesn’t update the selection-mode attribute
If you call `grid.selection.selectAll()`, the `selection-mode` attribute on the element is not updated to “all”. The `grid.selection.mode` property returns “all” correctly.

At the same time,  check that setting the select.mode property reflects correctly, and that selection.clear() correctly sets the attribute and property from "all" to "multi" (when in single-select mode, the mode doesn’t change when clear() is called).

manolo:
Fixes #247

This approach breaks selectAll() in the case of selection model == single, in the sense that it does nothing instead of clearing selections.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/268)


Saulis:
Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

_[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 370 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268#-K3oDRvzuKaqsQ8qa66P:-K3oDRvzuKaqsQ8qa66Q:1987655107) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b7e85f64b3d4cc1abf792d15387e070f63275b35/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L370)):_
I would refactor this into:

````javascript
getSelectionModel().clear();

if (getSelectionModel().getMode() == ALL) {
  setSelectionMode("multi", true);
}
````

To make it more explicit that `clear()` is the primary functionality in this method and changing the selection mode to "multi" is a side-effect. Also, I don't think we should use the `setSelectionMode` side-effect of clearing the selection when mode changes here, but instead explicitly always call `clear()` for all modes.

---

_[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 379 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268#-K3oC0GVHjmUJf7AlRBF:-K3oC0GWPzzHlXQDSBw9:1033534576) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b7e85f64b3d4cc1abf792d15387e070f63275b35/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L379)):_
I would remove the if-clause here and just call `selectAll()` on all selection modes. Single and Disabled implementations have a NO-OP in place. And I would call `selectAll()` explicitly first.

---

_[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 388 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268#-K3o9viHseFq_nxV88XM:-K3o9viHseFq_nxV88XN:-1951125649) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b7e85f64b3d4cc1abf792d15387e070f63275b35/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L388)):_
Why do we have the `force` argument? From what I can tell it's `false` only when selection mode is changed using the JS API. Sounds to me like we should rather be looking at if the selection mode has actually changed.

---

_[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 402 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268#-K3oBcjrBmgBjWByTx6f:-K3oBcjrBmgBjWByTx6g:1302998597) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b7e85f64b3d4cc1abf792d15387e070f63275b35/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L402)):_
I would move this to in `selectAll()` method to reduce complexity and side-effects on this method.

---

_[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 409 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268#-K3oCfQqEDaEr8F-2qo9:-K3oCfQqEDaEr8F-2qoA:-2120688388) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b7e85f64b3d4cc1abf792d15387e070f63275b35/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L409)):_
I guess this should fire only if the mode actually changes?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268)_

Saulis:
Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

_[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 379 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268#-K3oC0GVHjmUJf7AlRBF:-K3oEum1InezD7IaPhd0:1663336421) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b7e85f64b3d4cc1abf792d15387e070f63275b35/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L379)):_
Discard what I said about the order, changing the mode afterwards would clear the selections.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268)_

tomivirkki:
_[vaadin-grid.html, line 873 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268#-K3oKMT39DTb41pgxySr:-K3oKMT39DTb41pgxySs:-1958684042) ([raw file](https://github.com/vaadin/vaadin-grid/blob/0f53a3d02a0c41680e7fc985efd9de23814340b5/vaadin-grid.html#L873)):_
Distributing the selection api to different locations sounds a bit scary. Why not keeping them in the same place and using Grid events to message GridElement just like before?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/268)_

PR #270, opened 2015-11-24
Fix selection checkboxes to use --default-primary-color

### LINKED ISSUE #240: Rudimentary style customizability
We should offer the following styling options to make vaadin-grid a little easier to take into use:

- The blue highlight color should use `--default-primary-color` from paper-styles, and fallback to the same fixed color value if the custom property is not available/defined
- The row height should be exposed as a custom property, `--vaadin-grid-row-height`
- The font-family should be inherited from the parent element/page

All custom properties should be documented, and there should be a code example how to change the primary color and the row height.

tehapo:
Fixes #240

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/270)

### DIFF
--- a/demo/sizing.html
+++ b/demo/sizing.html
@@ -221,7 +221,7 @@ layout: page
       }
     </style>
     <code-example>
-      <vaadin-grid visible-rows=5 class="themed" demo hidden>
+      <vaadin-grid visible-rows=5 selection-mode="multi" class="themed" demo hidden>
         <table>
           <colgroup>
             <col>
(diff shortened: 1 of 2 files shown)


jounik:
I think the `var(...)` values should specify the fallback value also, like we do for the row focus styles.

tehapo:
Right, of course. :+1: Fixed.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/270#-:-K3tR65Nmt0Ky-WbXhDI:-247676491)_

Saulis:
Reviewed 2 of 2 files at r1.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/270)_

PR #312, opened 2016-01-15
Fix details not shown with shadow dom.

### LINKED ISSUE #194: Details row is not visible when using Shadow DOM
Copied the full source of the [Row Details](https://cdn.vaadin.com/vaadin-components/latest/vaadin-grid/demo/details.html) example and tried running it with ```?dom=shadow``` enabled. The result is that Grid reserves correct height of free space for the details, but nothing is visible on screen. See the attached screenshot.

screen shot 2015-10-05 at 11 46 02

manolo:
Fixes #194

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/312)

### DIFF
--- a/vaadin-grid.html
+++ b/vaadin-grid.html
@@ -239,6 +239,10 @@ See the [demo](demo/index.html) for use case examples.
     opacity: 1;
   }
 
+  .vaadin-grid-spacer td {
+    opacity: 1;
+  }
+
   /* Row styles */
   .vaadin-grid-row {
     box-sizing: border-box;


tomivirkki:
Was there something wrong with the originally proposed solution?

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 233 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-K84bB9ed8o6RC1ON1CB:-K84bB9ed8o6RC1ON1CC:113297663) ([raw file](https://github.com/vaadin/vaadin-grid/blob/cf8dedaf1e0796c98e3c58e5f00ac5aba7a62db3/vaadin-grid.html#L233)):*
There's no reason to have this change, since the selector beneath overrides it anyway.

---

*[vaadin-grid.html, line 237 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-K84bK0-_IE4X7ZLGWnj:-K84bK0-_IE4X7ZLGWnk:-1833951254) ([raw file](https://github.com/vaadin/vaadin-grid/blob/cf8dedaf1e0796c98e3c58e5f00ac5aba7a62db3/vaadin-grid.html#L237)):*
Is there a reason to add 1px bottom border to the td (to share all the styles that are meant for a data row)?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-:-K84eHL4e2eV05FO6lqu:-364716650)*

jounik:
Also, as a general CSS formatting rule, always place every individual selector on it’s own line.

manolo:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid.html, line 233 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-K84bB9ed8o6RC1ON1CB:-K84nR0HQ4s28O4TNGj2:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/cf8dedaf1e0796c98e3c58e5f00ac5aba7a62db3/vaadin-grid.html#L233)):*
Done.

---

*[vaadin-grid.html, line 237 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-K84bK0-_IE4X7ZLGWnj:-K84nRLHtGmw513OBg34:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/cf8dedaf1e0796c98e3c58e5f00ac5aba7a62db3/vaadin-grid.html#L237)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312)*

manolo:
@tomivirkki :  details `<td>` or it's parent `<tr>` is never classed, so there is no way to know whether it has data. Therefore it's better just fix its to 1. The only issue I can see is  that there is no transition when the row get data, but the details is not shown until its reference row has data and the details generator is called.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-:-K85gAF2nxjDISSkanoj:139742290)*

tomivirkki:
The details <td> doesn't have "vaadin-grid-cell" class as you said so the css blocks that hide the td's could be easily changed to target only "td.vaadin-grid-cell" as was suggested. The current approach adds unnecessary lines to unhide something that wasn't meant to be hidden.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-:-K8ImRnoiHaVeYTy0saW:1902884186)*

tomivirkki:
Actually just ".vaadin-grid-cell" instead of "td.vaadin-grid-cell" is enough

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-grid/312#-:-K8Io9nRfbK-2u-b9oeW:1183263903)*

PR #362, opened 2016-05-03
Grid fires an event when it's light dom table config has been applied

tomivirkki:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/362)


tomivirkki:
*[vaadin-grid.html, line 1132 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGpNA49I1lRjxjfQamj:-KGpNA49I1lRjxjfQamk:1732938600) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/vaadin-grid.html#L1132)):*
Now that the nature of the event has slightly changed (it's fired for every change caught by the observeNodes), should it be renamed to something else or is "grid-ready" still ok?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

manolo:
Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-grid.html, line 1132 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGpNA49I1lRjxjfQamj:-KGsKx43w2zXipV8L8pr:-76219141) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/vaadin-grid.html#L1132)):*
To me grid-ready means once, just when it is initialised.
This should be renamed to other thing else like grid-rendered grid-done?
I don't  see the need of this event  having the .then() method public, why to complicate the API so? 
I would add this for the future if someone ask for it.
Right now we only need one event fired just once.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

manolo:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[test/grid-rendering-light-dom.html, line 190 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGsMcEsdf5uymXL66Oh:-KGsMcEsdf5uymXL66Oi:-797334899) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/test/grid-rendering-light-dom.html#L190)):*
Should be fired not when the grid is configured, but when the grid is actually rendered and everything in place, instead of .sync you should use .then

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

tomivirkki:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 1132 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGpNA49I1lRjxjfQamj:-KGuOhpplnMxrIVRizXa:269067822) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/vaadin-grid.html#L1132)):*
> To me grid-ready means once, just when it is initialised.
Yeas, that's why I'm suggesting the name change

> This should be renamed to other thing else like grid-rendered grid-done?
-rendered relates to UI rendering so it would sound strange here. -done doesn't differ from -ready in any way, just different wording for same thing. Maybe -configured.

> I don't see the need of this event having the .then() method public, why to complicate the API so?
I'm not following. If you're referring to the 'this' detail, It's needed in Angular at least so you don't need to maintain a reference to the dom element in component class.

> Right now we only need one event fired just once.
From Polymer developer's point of view, adding an event that fires just once when grid is ready is quite redundant. They already have direct access to the element and can use then() -api just as well. How do we justify adding the event for other than Angular 2 devs?

---

*[test/grid-rendering-light-dom.html, line 190 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGsMcEsdf5uymXL66Oh:-KGuQcH-WAeZeGeT2I7A:-407194797) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/test/grid-rendering-light-dom.html#L190)):*
Shouldn't one be able to configure grid's columns even if it's not visible yet?
If we change the spec like that then "grid-ready" sounds even stranger for the event name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

manolo:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[test/grid-rendering-light-dom.html, line 190 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGsMcEsdf5uymXL66Oh:-KGvBk67s5B9kpLANrF-:1182137008) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/test/grid-rendering-light-dom.html#L190)):*
Tested this code and works ok. So as we can configure columns before it's attached. 

ready: function() {
    .... _grid.init

      this.then(function() {
        this.fire('grid-ready', this);
      }.bind(this));

   ....
}

I suggested attach: because there were a grid.then there. But you are right, people would want to configure the grid when its ready.

I think having this event in the ready block makes more  sense to name it grid-ready 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

tomivirkki:
*[test/grid-rendering-light-dom.html, line 190 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGsMcEsdf5uymXL66Oh:-KGvMWBItyoLzeIcXo9u:-1923748996) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/test/grid-rendering-light-dom.html#L190)):*
I'm fine with this suggestion

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

manolo:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 1132 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGpNA49I1lRjxjfQamj:-KGvMb5JMfrpbEeG5Uo0:-1234144413) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/vaadin-grid.html#L1132)):*
> I'm not following. If you're referring to the 'this' detail, It's needed in Angular at least so you don't need to maintain a reference to the dom element in component class.

You don't maintain that reference because angular does it for you via the the native element getter.

>  They already have direct access to the element and can use then()

IIRC, we added the event in grid.directive in order to make easier to ng developers know when the grid is ready. It seemed easier that they listen to an event instead of inspect the api looking for then().
When removing grid directive, my first approach was to provide this feature into the generic directive, so as it's able to dynamically inspect when an element needs a while to be ready, but seemed very hacky

So there are three options, that the user call then() method so we don't modify the grid api, generic directive considers these corner cases, or fire a ready event, I don't have 

I don't have any preference, but I see that firing dom change events  is out of this scope

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

manolo:
Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[test/grid-rendering-light-dom.html, line 190 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362#-KGsMcEsdf5uymXL66Oh:-KGvRfDVXjxgwOZ-4vGo:1870623875) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f5706211ec1bc21074cb6cab9b26c76d37d01bf9/test/grid-rendering-light-dom.html#L190)):*
Yes I think this could be a simple solution and they don't have to deal with then() although they can use anytime.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/362)*

tomivirkki:
Closing this PR. We can emphasize using @ViewChild with then() in Angular 2 apps instead.

PR #394, opened 2016-06-23
Column reorder

djbowman1:
Added functionality to webcomponent, vaadin-grid, that lets user enable column reordering by drag and drop. 
 - Methods added in webcomponent to enable/disable and also in gridElement and violatedGrid now initializes columnReorderingAllowed to false.
 - Added columnReorder event handler to gridElement, made gridElement now update its locally stored cols from grid whenever columnReordering occurs.Also gridElements now fires html event on columnReorder, which webcomponent listens for to update its local copy of columns.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/394)


jstanley1730:
awesome!

vaadin-grid.html
```diff
@@ -1035,7 +1048,21 @@
...
+      /*
+      * Boolean property to set whether column reordering is allowed by the grid.
+      *
+      * @property {boolean} columnReorderingAllowed
+      * @default false
+      * @type {boolean}
+      */
+      columnReorderingAllowed: {
+        type: Boolean,
+        default: false,
```
billy-hardy: do you mean `value: false`?

manolo:
We like to have tests at least per each feature added to the component. 
The ideal would be if you add it, but if you are not familiar and you consider it could be complex we could add it latter or before merging this.

---

Review status: 0 of 4 files reviewed at latest revision, 13 unresolved discussions.

---

*[vaadin-grid.html, line 1062 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKyZpPnLNTzmT0ZP93p-r1-1062:-KKzREwcB8Byut43BUyc:1414680644) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/vaadin-grid.html#L1062)):*
<details><summary><i>Previously, billy-hardy (Billy Hardy) wrote…</i></summary>
> do you mean `value: false`?
</details>
Yes it should be value, or just remove the line.

---

*[vaadin-grid.html, line 1101 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzP7fGUhKmkL6NnDng:-KKzP7fGUhKmkL6NnDnh:-38546478) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/vaadin-grid.html#L1101)):*
> ```HTML
>     ],
> 
>     _onColumnOrderChanged: function(){
> ```

we are removing the 'on' prefix for these methods, I have no preferences, but I think it's better to keep aligned and just name it  `_columnOrcerChanged`

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 114 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPav5dmYH76KurIre:-KKzPav5dmYH76KurIrf:1604809463) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L114)):*
> ```Java
>         grid.setStylePrimaryName("vaadin-grid style-scope vaadin-grid");
>     }
>     
> ```

We remove empty spaces at the end of any line, so keep this line empty.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 143 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPmIN4av2HRpP7Hv1:-KKzPmIN4av2HRpP7Hv2:-1242518158) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L143)):*
> ```Java
>         return order;
>     }
>     
> ```

spaces

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 147 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPnrW58Id5ikN1-j3:-KKzPnrW58Id5ikN1-j4:1180884872) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L147)):*
> ```Java
>         List<GridColumn> dataColumns = getDataColumns();
>         if (dataColumns.size() == 0)
>         	return;
> ```

We use 4 spaces instead of tabs

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 148 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPqaIBpRrR4P1bMyO:-KKzPqaIBpRrR4P1bMyP:-1242518158) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L148)):*
> ```Java
>         if (dataColumns.size() == 0)
>         	return;
>         
> ```

spaces

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 149 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPywJ7ocAljP5pHzV:-KKzPywKuEPTSfjLPYLE:-1348441369) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L149)):*
> ```Java
>         	return;
>         
>         for (int i = cols.size(); i>0;i--){
> ```

the format is incorrect  should be  `;  i > 0; i--` note spaces

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 150 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPs7lWUwIUpMCFjnq:-KKzPs7lWUwIUpMCFjnr:678051331) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L150)):*
> ```Java
>         
>         for (int i = cols.size(); i>0;i--){
>         	cols.remove(cols.get(0));
> ```

tabs

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 153 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPswkr9uqwXigpBUq:-KKzPswkr9uqwXigpBUr:678051331) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L153)):*
> ```Java
>         }
>         for (int i = 0; i<dataColumns.size();i++){
>         	cols.add(dataColumns.get(i).getJsColumn());
> ```

tabs

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 608 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPuoIH_V56f-nNzM7:-KKzPuoIH_V56f-nNzM8:678051331) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L608)):*
> ```Java
>     @JsIgnore 
>     @Override
> 	public void onColumnReorder(ColumnReorderEvent<Object> event) {
> ```

tabs

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 77 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzQBs9KcgYc1-dQWlT:-KKzQBs9KcgYc1-dQWlU:-1541054071) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L77)):*
> ```Java
>             }
>         }
>         super.setColumnReorderingAllowed(false);
> ```

I think by default it's already false so no need to call this method here

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 175 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzQT-T951sNwf7dA7u:-KKzQT-UyrryD33kYQec:1694009400) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L175)):*
> ```Java
>             super.onBrowserEvent(event);
>         }
>         if (targetElement == null){
> ```

I feel like this check is not necessary, so remove it unless you have a good reason for this.
In case of we want to throw an exception I would do at the beginning of the method.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 176 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzQIJZMNs0TAMpXA4n:-KKzQIJZMNs0TAMpXA4o:-455723231) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L176)):*
> ```Java
>         }
>         if (targetElement == null){
>         	throw new NullPointerException();
> ```

space

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-:-KKzQn_ujcP5XngexXiV:-1412151912)*

djbowman1:
@manolo Made all the requested changes, however I've had little experience testing in java, could you give some guidance as to how one would go about testing with what I've added?

djbowman1:
Review status: 0 of 4 files reviewed at latest revision, 13 unresolved discussions.

---

*[vaadin-grid.html, line 1062 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKyZpPnLNTzmT0ZP93p-r1-1062:-KL1SaDSu6_urah3hQLS:-1760911079) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/vaadin-grid.html#L1062)):*
> Yes it should be value, or just remove the line.
</details>
Removed the line, so its now defaulting to null, which it actually was before because the "default" actually did nothing

---

*[vaadin-grid.html, line 1101 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzP7fGUhKmkL6NnDng:-KL1SgqKV6WDVggkoW5e:1761390629) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/vaadin-grid.html#L1101)):*
> we are removing the 'on' prefix for these methods, I have no preferences, but I think it's better to keep aligned and just name it  `_columnOrcerChanged`
</details>
Done. Method name changed

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 114 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPav5dmYH76KurIre:-KL1S-LqNQqUwrfE6nxp:1706342915) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L114)):*
> We remove empty spaces at the end of any line, so keep this line empty.
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 143 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPmIN4av2HRpP7Hv1:-KL1S4EcIEGhy447B8nR:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L143)):*
> spaces
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 147 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPnrW58Id5ikN1-j3:-KL1S4rB6gvYsuM28jjj:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L147)):*
> We use 4 spaces instead of tabs
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 148 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPqaIBpRrR4P1bMyO:-KL1S5LBWycyMPCQY74K:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L148)):*
> spaces
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 149 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPywJ7ocAljP5pHzV:-KL1S5ri4Ur6ELiw54uC:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L149)):*
> the format is incorrect  should be  `;  i > 0; i--` note spaces
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 150 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPs7lWUwIUpMCFjnq:-KL1S7ZpSDciU4oI6Yy1:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L150)):*
> tabs
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 153 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPswkr9uqwXigpBUq:-KL1S82147cmQcu3jkDA:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L153)):*
> tabs
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 608 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzPuoIH_V56f-nNzM7:-KL1S8lgsEtBT6z21yla:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L608)):*
> tabs
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 77 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzQBs9KcgYc1-dQWlT:-KL1SAmwQ4mDGks-xCbK:1561207366) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L77)):*
> I think by default it's already false so no need to call this method here
</details>
Done. Method call removed, still functions correctly, this was mostly just there as testing that this worked early on.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 175 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzQT-T951sNwf7dA7u:-KL1SJA91xC6LgXO2SKF:539887167) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L175)):*
> I feel like this check is not necessary, so remove it unless you have a good reason for this.
> In case of we want to throw an exception I would do at the beginning of the method.
</details>
Done. This check was unnecessary, was getting a weird error before and this fixed it, but this was because of wrong/weird way of doing something that was changed.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 176 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KKzQIJZMNs0TAMpXA4n:-KL1SYtvxV1kwJRi88FA:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/836c712472705e4ff115ec663a095a07c506ff4a/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L176)):*
> space
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

jstanley1730:
@manolo what is status of review of this pull request, seems all issues have been resolved except for test, could you please provide guidance on what is desired to satisfy the testing requirements

manolo:
If you see the test folder,  we are not doing any java testing, instead we do everything  with Polymer Web Component Tester (WCT) in javascript .  There are a lot of tests you can take a reference. Then running polyserve and navigating to the test folder you can test in different browsers.

Another thing missing is that you could update some of the column demos to configure it for column reordering, writing a line with the feature.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-:-KL2KakniE5HB8IXMdVy:-856071431)*

djbowman1:
We actually just found a bug in grid that is preventing the dragging to work at all times, I will be making a branch/pull request on vaadin before I can finish this ticket. I will also be writing a wct test and adding a column-reordering demo.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-:-KLbojsxvpKaSCdC1CPE:-1197456649)*

djbowman1:
@manolo I've added in tests and a small demo showcasing column reordering, the only thing thats left that needs updated is the changes I've made to the FW grid that needs pulled over into this. How would one go about getting that done?

manolo:
FW grid is merged. They should to make a release soon, but it's not a blocker for this PR.
There is some comments here, almost small syntax issues because we are strict with conventions.
I think fixing them we are ready to merge it. Since this is a new feature it would not be included in 1.1.1 but in 1.2.0 

---

Review status: 0 of 7 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[demo/columns.html, line 270 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN1z5Q4coZiKYbvinel:-KN1z5Q5eGE43kZ7qgbZ:2052773000) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L270)):*
> ```HTML
>   <section>
>     <h3>Column Reordering</h3>
>     <p>Columns can reordered by user drag and drop. Column reordering will respect frozen columns and is
> ```

The comment is not clear enough, could you change to something like?: 

Non frozen columns can be reordered by dragging header cells. By default reordering is disabled, thus  
you have either to set the `columnReorderingAllowed` property, or to add the `column-reordering-allowed` attribute to enable it.

---

*[demo/columns.html, line 293 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2-pJSzRWL5XGXW3kz:-KN2-pJSzRWL5XGXW3l-:-2100341039) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L293)):*
> ```HTML
>         </table>
>       </vaadin-grid>
>       <code demo-var="grid">
> ```

I would remove all the JS code block and add the attribute column-reordering-allowed to the element. Normally people prefer declarative language for configuration.
Also I would add one frozen column so as the user can experiment with the complete feature.
Your html code should look like 
` <vaadin-grid demo frozen-columns="1" column-reordering-allowed >

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 100 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21_bjyR1sJvfoJWpm:-KN21_bjyR1sJvfoJWpn:-1211037758) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L100)):*
> ```Java
>     private static final String SELECTION_MODE_CHANGED_EVENT = "selection-mode-changed";
>     private static final String COLUMN_ORDER_CHANGED_EVENT = "column-order-changed";
>     
> ```

extra spaces

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 146 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN20eTzIBQ9KpIUdT1N:-KN20eTzIBQ9KpIUdT1O:497796726) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L146)):*
> ```Java
>     private void updateOrder(){
>         List<GridColumn> dataColumns = getDataColumns();
>         if (dataColumns.size() == 0)
> ```

Add curly brackets, it's our convention.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 148 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2188YVMDvz90KV4CA:-KN2188YVMDvz90KV4CB:55052972) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L148)):*
> ```Java
>         if (dataColumns.size() == 0)
>             return;
>         for (int i = cols.size(); i > 0; i--){
> ```

Add space previous to the curly bracket

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 151 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21BQuuDU_7X9yO7l5:-KN21BQuuDU_7X9yO7l6:-1406215184) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L151)):*
> ```Java
>             cols.remove(cols.get(0));
>         }
>         for (int i = 0; i<dataColumns.size();i++){
> ```

The syntax here should be  
`for (int i = 0; i < dataColumns.size(); i++) {`

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 607 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21ThM3zijXxTzUH9G:-KN21ThM3zijXxTzUH9H:331484932) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L607)):*
> ```Java
>     @JsIgnore 
>     @Override
> 	public void onColumnReorder(ColumnReorderEvent<Object> event) {
> ```

You are using tabs instead of spaces

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 610 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21WcxwukSYJyIxUct:-KN21WcxwukSYJyIxUcu:678051331) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L610)):*
> ```Java
>         updateOrder();
>         triggerEvent(COLUMN_ORDER_CHANGED_EVENT);
> 	}
> ```

tabs

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 611 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21Xl_IvR41K8W3s_d:-KN21Xl_IvR41K8W3s_e:-1211037758) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L611)):*
> ```Java
>         triggerEvent(COLUMN_ORDER_CHANGED_EVENT);
> 	}
>     
> ```

extra spaces

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-:-KN21nyaHfp4mCEqNHC2:1483012292)*

platosha:
Reviewed 4 of 4 files at r2, 2 of 3 files at r3, 1 of 2 files at r4.
Review status: 6 of 7 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 121 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2Fb9MyeBkJioZDNe4:-KN2Fb9MyeBkJioZDNe5:1164234594) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L121)):*
> ```HTML
> `--vaadin-grid-row-cell` | Mixin which applies to the cell elements of the table | {}
> `--header-drag-table-background-color` | Mixin which applies to the background of header drag table | `#EEEEEE`
> `--header-drop-marker-color` | Mixin which applies to the color of the drop marker | `#03A9F4`
> ```

`--header-drag-table-background-color` and `--header-drop-marker-color` are described as mixins, though they are applied as not mixins, but properties, currently.

Vaadin Grid doesn’t have custom CSS properties for colors, as for now, but reuses ones from `paper-styles/default-theme.html`. I would not like to introduce those in this PR as well. Please remove the grid color properties. Use `--primary-color` and other colors of the default theme instead, where appropriate.

It would be nice to also have true custom CSS mixins instead of these color properties, similar to the`--vaadin-grid-row-cell` mixin that is applied for all cells:
```css
  .vaadin-grid-header th,
  .vaadin-grid-body td,
  .vaadin-grid-footer td {
    ...
    @apply(--vaadin-grid-row-cell);
  }
```

If you add style mixins, please name them with a prefix, such as: `--vaadin-grid-<some>-row-cell`. It is important, because they can inherit values from the global scope.

Or we can add them back in the separate PR to speed up this review.

---

*[vaadin-grid.html, line 668 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2JN-63ogIPzckN3J7:-KN2JN-7UtEFSVBz41-2:1854541912) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L668)):*
> ```HTML
>   }
> 
>  :host ::content .header-drag-table .vaadin-grid-cell {
> ```

`:host ::content` means apply the styles to the light DOM contents, but `.header-drag-table` is internal local DOM part of the grid. When Polymer uses shadow DOM mode (append `?dom=polymer` to the URL to enable), these styles will not apply to the grid internals.

Please remove `:host ::content` from the internal styles.

---

*[vaadin-grid.html, line 674 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2K0b3tumV74dDCpzS:-KN2K0b3tumV74dDCpzT:1106009288) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L674)):*
> ```HTML
>  }
> 
>  :host ::content .header-drag-table .vaadin-grid-drop-marker {
> ```

Please remove `:host ::content` from the internal styles.

---

*[vaadin-grid.html, line 1355 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2Nu653rNZIm7jVJjE:-KN2Nu66E5cs7OFHzkrO:-2065777948) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L1355)):*
> ```HTML
>     getItem: function(rowIndex, callback, onlyCached) {
>       this._grid.getItem(rowIndex, callback, onlyCached);
>     },
> ```

Unnecessary comma and the empty line below.

---

*[test/grid-column-reordering.html, line 35 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2R7r2piDsKH6aa55H:-KN2R7r2piDsKH6aa55I:849045272) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L35)):*
> ```HTML
> 
>     it('columnReorderingAllowed changes listener test', function() {
> ```

Unnecessary empty line.

---

*[test/grid-column-reordering.html, line 36 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2PBF0gEYZey2s4pO1:-KN2PBF0gEYZey2s4pO2:-731633365) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L36)):*
> ```HTML
>     it('columnReorderingAllowed changes listener test', function() {
> 
>       var colReorderingAllowedChangedSpy = sinon.spy(grid, "_columnReorderingAllowedChanged");
> ```

It is probably redundant to spy for the `_columnReorderingAllowedChanged` observer calls.

IMO, it is only important to test, that the `_grid.setColumnReorderingAllowed` is called with the new value, the listener method is just an implementation detail of that. If the listener is broken, the `_grid.setColumnReorderingAllowed` call test should also catch it.

The listener call when the property value changes is a feature of the Polymer library, not of Vaadin Grid, so it is out of our scope for testing.

---

*[test/grid-column-reordering.html, line 53 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2R0M32xW3bpR8Eio3:-KN2R0M4WTXSffFlH3Iy:849045272) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L53)):*
> ```HTML
> 
>     it('column-order-changed listener test', function(){
> ```

Unnecessary empty line.

---

*[test/grid-column-reordering.html, line 54 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2TB9cdmhov5uoUJcX:-KN2TB9cdmhov5uoUJcY:-1264302301) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L54)):*
> ```HTML
>     it('column-order-changed listener test', function(){
> 
>       var colOrderChangedSpy = sinon.spy(grid, '_columnOrderChanged');
> ```

Testing the listener might be redundant here as well.

---

*[test/grid-column-reordering.html, line 59 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2QT4_hYhE6hOS4Mh0:-KN2QT4_hYhE6hOS4Mh1:893159327) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L59)):*
> ```HTML
>       var colsBeforeUpdate = grid.columns;
>       grid.fire('column-order-changed');
>       expect(colsBeforeUpdate).to.equal(grid.columns);
> ```

Mind that this will not catch any mutations of the `grid.columns` object, as long as the reference stays the same.

---

*[test/grid-column-reordering.html, line 63 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2RimuCZCcuWfQcs3Z:-KN2RimuCZCcuWfQcs3_:1737719278) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L63)):*
> ```HTML
>       expect(getColumnsSpy.calledOnce).to.be.true;
>     });
> ```

Looks like the actual tests for column reordering are missing. I mean, testing for whether the order of columns changes after dragging something. 

The `MockInteractions.track(target, dx, dy, steps)` method from `iron-test-helpers` might help.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

manolo:
Review status: 6 of 7 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 121 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2Fb9MyeBkJioZDNe4:-KN2ZXwxcetajn8nZivv:1969646618) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L121)):*
> `--header-drag-table-background-color` and `--header-drop-marker-color` are described as mixins, though they are applied as not mixins, but properties, currently.
> 
> Vaadin Grid doesn’t have custom CSS properties for colors, as for now, but reuses ones from `paper-styles/default-theme.html`. I would not like to introduce those in this PR as well. Please remove the grid color properties. Use `--primary-color` and other colors of the default theme instead, where appropriate.
> 
> It would be nice to also have true custom CSS mixins instead of these color properties, similar to the`--vaadin-grid-row-cell` mixin that is applied for all cells:
> ```css
>   .vaadin-grid-header th,
>   .vaadin-grid-body td,
>   .vaadin-grid-footer td {
>     ...
>     @apply(--vaadin-grid-row-cell);
>   }
> ```
> 
> If you add style mixins, please name them with a prefix, such as: `--vaadin-grid-<some>-row-cell`. It is important, because they can inherit values from the global scope.
> 
> Or we can add them back in the separate PR to speed up this review.
</details>
- +1 to do the mixins in a separated PR.
- for colors it sound reasonable to remove that properties and
  - you use hardcoded colors like we do with vaadin-grid-row-focused, (we should fix everything in a separated PR though)
  - use the same properties that polymer uses, for the marker color: var(--primary-color, #00B4F0) and for the background: var(--primary-background-color, #EEEEEE)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

djbowman1:
Review status: 3 of 7 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 121 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2Fb9MyeBkJioZDNe4:-KN3f0Ta_MzheBcXJra1:1752505579) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L121)):*
> - +1 to do the mixins in a separated PR.
> - for colors it sound reasonable to remove that properties and
>   - you use hardcoded colors like we do with vaadin-grid-row-focused, (we should fix everything in a separated PR though)
>   - use the same properties that polymer uses, for the marker color: var(--primary-color, #00B4F0) and for the background: var(--primary-background-color, #EEEEEE)
</details>
changed to use the vars pointed out, mixins removed

---

*[vaadin-grid.html, line 1355 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2Nu653rNZIm7jVJjE:-KN82Chmr829bz5K7ZNl:-1054325829) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L1355)):*
> Unnecessary comma and the empty line below.
</details>
removed

---

*[demo/columns.html, line 270 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN1z5Q4coZiKYbvinel:-KN82KdTyfDzbGSlYMuK:-1274186194) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L270)):*
> The comment is not clear enough, could you change to something like?: 
> 
> Non frozen columns can be reordered by dragging header cells. By default reordering is disabled, thus  
> you have either to set the `columnReorderingAllowed` property, or to add the `column-reordering-allowed` attribute to enable it.
> 
</details>
Comment changed

---

*[demo/columns.html, line 293 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2-pJSzRWL5XGXW3kz:-KN8ZQm6vWF1uXzTriY7:-2111397126) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L293)):*
> I would remove all the JS code block and add the attribute column-reordering-allowed to the element. Normally people prefer declarative language for configuration.
> Also I would add one frozen column so as the user can experiment with the complete feature.
> Your html code should look like 
> ` <vaadin-grid demo frozen-columns="1" column-reordering-allowed >
> 
</details>
js code block removed, attribute set in markup, also frozen column added to demo

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 100 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21_bjyR1sJvfoJWpm:-KN8ZeAQ7C4acIflDcVi:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L100)):*
> extra spaces
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 146 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN20eTzIBQ9KpIUdT1N:-KN8ZhkSR7I_zZVzEtao:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L146)):*
> Add curly brackets, it's our convention.
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 148 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2188YVMDvz90KV4CA:-KN8ZiyQltTK4RLf7924:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L148)):*
> Add space previous to the curly bracket
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 151 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21BQuuDU_7X9yO7l5:-KN8ZoFi_xYnW139_lko:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L151)):*
> The syntax here should be  
> `for (int i = 0; i < dataColumns.size(); i++) {`
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 607 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21ThM3zijXxTzUH9G:-KN8ZpojxmHeDMIDNHBv:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L607)):*
> You are using tabs instead of spaces
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 610 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21WcxwukSYJyIxUct:-KN8Zxvsd2_aqXSAbbyi:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L610)):*
> tabs
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/GridElement.java, line 611 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN21Xl_IvR41K8W3s_d:-KN8_01iHcxBg4m_1iN6:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/java/src/main/java/com/vaadin/elements/grid/GridElement.java#L611)):*
> extra spaces
</details>
Done.

---

*[test/grid-column-reordering.html, line 35 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2R7r2piDsKH6aa55H:-KN8_FuJYbyV_7CD2YTk:-2005434329) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L35)):*
> Unnecessary empty line.
</details>
Done. removed

---

*[test/grid-column-reordering.html, line 36 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2PBF0gEYZey2s4pO1:-KNHaCs9vvfhQUDAPjZy:203803171) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L36)):*
> It is probably redundant to spy for the `_columnReorderingAllowedChanged` observer calls.
> 
> IMO, it is only important to test, that the `_grid.setColumnReorderingAllowed` is called with the new value, the listener method is just an implementation detail of that. If the listener is broken, the `_grid.setColumnReorderingAllowed` call test should also catch it.
> 
> The listener call when the property value changes is a feature of the Polymer library, not of Vaadin Grid, so it is out of our scope for testing.
</details>
Done. Spy removed

---

*[test/grid-column-reordering.html, line 53 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2R0M32xW3bpR8Eio3:-KN8_eTceRVnYD5_7Isg:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L53)):*
> Unnecessary empty line.
</details>
Done.

---

*[test/grid-column-reordering.html, line 54 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2TB9cdmhov5uoUJcX:-KNHaMSC_u2A5Ql-gDSc:-340781049) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L54)):*
> Testing the listener might be redundant here as well.
</details>
Done. Removed

---

*[test/grid-column-reordering.html, line 59 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2QT4_hYhE6hOS4Mh0:-KN8_orWRzY9y79hbly0:1772071634) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L59)):*
> Mind that this will not catch any mutations of the `grid.columns` object, as long as the reference stays the same.
</details>
didnt even realize that i was doing a copy by reference, changed

---

*[test/grid-column-reordering.html, line 63 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2RimuCZCcuWfQcs3Z:-KNHvLQeK76YJGHFEukX:-132098397) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/test/grid-column-reordering.html#L63)):*
> Looks like the actual tests for column reordering are missing. I mean, testing for whether the order of columns changes after dragging something. 
> 
> The `MockInteractions.track(target, dx, dy, steps)` method from `iron-test-helpers` might help.
</details>
Extra tests added to test functionality while using mockInteractions to drag elements, both with and without frozen columns

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

manolo:
Fix selectors, and a couple of comments, otherwise looks good to me.

---

Reviewed 1 of 2 files at r4, 4 of 4 files at r5.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 121 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2Fb9MyeBkJioZDNe4:-KNVjkFi6_PQvaMCaEoo:-242550700) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L121)):*
<details><summary><i>Previously, djbowman1 wrote…</i></summary>
> changed to use the vars pointed out, mixins removed
</details>
The last backquote in this line is not needed

---

*[vaadin-grid.html, line 668 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2JN-63ogIPzckN3J7:-KNVxwC9HsTAJWICPE84:-515972512) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L668)):*
> `:host ::content` means apply the styles to the light DOM contents, but `.header-drag-table` is internal local DOM part of the grid. When Polymer uses shadow DOM mode (append `?dom=polymer` to the URL to enable), these styles will not apply to the grid internals.
> 
> Please remove `:host ::content` from the internal styles.
</details>
So what Anton is saying is that these selectors do not work when using shadow dom, but probably you are not using it, and it's difficult to see the problem because our examples do not work in shadow dom because of the components we use for displaying them.

So change your selector by this block and should work. Note that first selector is for shady dom because grid does not use Polymer.dom api to manipulate elements, and second one is for shadow dom.

```
  :host ::content .dragged-column-header,
  .dragged-column-header {
    background-color: var(--primary-background-color, #EEEEEE);
    position: absolute;
    opacity: 0.5;
  }
```

Also note that we use two spaces for ident the css code, seems that you were using one. The block I'm pasting here should be ok.

---

*[vaadin-grid.html, line 674 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2K0b3tumV74dDCpzS:-KNVyVohYh17yEwi2lbl:-968701679) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L674)):*
> Please remove `:host ::content` from the internal styles.
</details>
Change this block by this:

```
  :host ::content .vaadin-grid-drop-marker,
  .vaadin-grid-drop-marker {
    background-color: var(--primary-color, #00B4F0);
    position: absolute;
    width: 3px;
  }
```

---

*[demo/columns.html, line 293 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2-pJSzRWL5XGXW3kz:-KNVj6HVFsgQwGcJmuOI:672045680) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L293)):*
<details><summary><i>Previously, djbowman1 wrote…</i></summary>
> js code block removed, attribute set in markup, also frozen column added to demo
</details>
I meant remove the  `<code>` block, it is useless now that there is no significant code in it. Also leaving this makes the demo page show an empty div without anything.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-:-KNVlZc_WMNtbyjnOhem:1301475583)*

manolo:
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[demo/columns.html, line 271 \[r5\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KNW2gFrRjGk9FA4IL9e:-KNW2gFrRjGk9FA4IL9f:1622428204) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f9eca7a63fe6db6235c396233a842a31725d7fb3/demo/columns.html#L271)):*
> ```HTML
>     <h3>Column Reordering</h3>
>     <p>Non frozen columns can be reordered by dragging header cells. By default reordering is disabled, thus
>       you have either to set the columnReorderingAllowed property, or to add the column-reordering-allowed attribute to enable it.</p>
> ```

I forgot this,  attributes and properties : `columnReorderingAllowed` and `column-reordering-allowed` could be wrapped with a `<code>..</code>` block like in other examples.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

djbowman1:
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-grid.html, line 121 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2Fb9MyeBkJioZDNe4:-KNbl4NFhkGYHG2cLmVj:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L121)):*
> The last backquote in this line is not needed
</details>
Done.

---

*[vaadin-grid.html, line 668 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2JN-63ogIPzckN3J7:-KNblVECBLKRe6CbnLHv:35349993) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L668)):*
> So what Anton is saying is that these selectors do not work when using shadow dom, but probably you are not using it, and it's difficult to see the problem because our examples do not work in shadow dom because of the components we use for displaying them.
> 
> So change your selector by this block and should work. Note that first selector is for shady dom because grid does not use Polymer.dom api to manipulate elements, and second one is for shadow dom.
> 
> ```
>   :host ::content .dragged-column-header,
>   .dragged-column-header {
>     background-color: var(--primary-background-color, #EEEEEE);
>     position: absolute;
>     opacity: 0.5;
>   }
> ```
> 
> Also note that we use two spaces for ident the css code, seems that you were using one. The block I'm pasting here should be ok.
</details>
Done. I cant believe the solution was this simple and we missed this, thanks for the fix

---

*[vaadin-grid.html, line 674 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2K0b3tumV74dDCpzS:-KNIVrvbJje8-mS7U1Ap:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/vaadin-grid.html#L674)):*
> Change this block by this:
> 
> ```
>   :host ::content .vaadin-grid-drop-marker,
>   .vaadin-grid-drop-marker {
>     background-color: var(--primary-color, #00B4F0);
>     position: absolute;
>     width: 3px;
>   }
> ```
</details>
Done.

---

*[demo/columns.html, line 293 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2-pJSzRWL5XGXW3kz:-KNbnwIcpMb33_ZiCfBp:137310669) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L293)):*
> I meant remove the  `<code>` block, it is useless now that there is no significant code in it. Also leaving this makes the demo page show an empty div without anything.
</details>
im going to ignore this one in favor of other comment saying to show markup/ 

---

*[demo/columns.html, line 271 \[r5\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KNW2gFrRjGk9FA4IL9e:-KNbt6duaKIdQRKo1Zqw:-1370773260) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f9eca7a63fe6db6235c396233a842a31725d7fb3/demo/columns.html#L271)):*
> I forgot this,  attributes and properties : `columnReorderingAllowed` and `column-reordering-allowed` could be wrapped with a `<code>..</code>` block like in other examples.
</details>
I have it set in the markup of the vaadin-grid, ```<vaadin-grid column-reordering-allowed>``` and also in a ```<code>...<code>``` block as ```columnReorderingAllowed = true``` so that in the demo we can show both ways of allowing it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

manolo:
Review status: 5 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[demo/columns.html, line 293 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2-pJSzRWL5XGXW3kz:-KNf9vFf7GrLsKZZNLJL:811905857) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L293)):*
<details><summary><i>Previously, djbowman1 wrote…</i></summary>
> im going to ignore this one in favor of other comment saying to show markup/ 
</details>
What I meant is you should remove everything from line 309 to 318 : 

```
   <code demo-var="grid">
   ...
  </code>
```

The confusion was that in two comments I was referring to code blocks, but they are unrelated comments.

---

*[demo/columns.html, line 271 \[r5\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KNW2gFrRjGk9FA4IL9e:-KNf9PN0cDqH5w2kuAl9:965955243) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f9eca7a63fe6db6235c396233a842a31725d7fb3/demo/columns.html#L271)):*
<details><summary><i>Previously, djbowman1 wrote…</i></summary>
> I have it set in the markup of the vaadin-grid, ```<vaadin-grid column-reordering-allowed>``` and also in a ```<code>...<code>``` block as ```columnReorderingAllowed = true``` so that in the demo we can show both ways of allowing it
</details>
I think you misunderstood what I was saying. What I meant is that line 271 should look like:

```
   <p>Non frozen columns can be reordered by dragging header cells. By default reordering is disabled, thus
      you have either to set the <code>columnReorderingAllowed</code> property, or to add the <code>column-reordering-allowed</code> attribute to enable it.</p>
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

manolo:
Review status: 5 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[demo/columns.html, line 293 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2-pJSzRWL5XGXW3kz:-KNfAkaM94BIf5ltFYbF:-1375454561) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L293)):*
> What I meant is you should remove everything from line 309 to 318 : 
> 
> ```
>    <code demo-var="grid">
>    ...
>   </code>
> ```
> 
> The confusion was that in two comments I was referring to code blocks, but they are unrelated comments.
</details>
Note that I think it's not necessary to show in code that we can do the same in declarative and imperative way, since we already mention it in the description of the feature

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

manolo:
Note that you have to rebase with master.

---

Review status: 5 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-:-KNfC2qsQSqpQbvWkdUQ:1452280144)*

platosha:
Reviewed 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

djbowman1:
Review status: 6 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[demo/columns.html, line 293 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KN2-pJSzRWL5XGXW3kz:-KNgUTjmMnkNYkm5JPwB:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/6fd3c1c0dd489b8d24fa8b21a78fe113b271a532/demo/columns.html#L293)):*
> Note that I think it's not necessary to show in code that we can do the same in declarative and imperative way, since we already mention it in the description of the feature
</details>
Done.

---

*[demo/columns.html, line 271 \[r5\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394#-KNW2gFrRjGk9FA4IL9e:-KNgUD6n9FrN5OaQ-ms3:-405757987) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f9eca7a63fe6db6235c396233a842a31725d7fb3/demo/columns.html#L271)):*
> I think you misunderstood what I was saying. What I meant is that line 271 should look like:
> 
> ```
>    <p>Non frozen columns can be reordered by dragging header cells. By default reordering is disabled, thus
>       you have either to set the <code>columnReorderingAllowed</code> property, or to add the <code>column-reordering-allowed</code> attribute to enable it.</p>
> ```
> 
> 
> 
</details>
Done. Sorry about misunderstanding with that

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

manolo:
Reviewed 4 of 4 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/394)*

manolo:
@djbowman1 I have merged your branch into master, and also a couple of minor commits to update the compiled js and fix syntax warnings.

Thank you very much for your contribution.

PR #397, opened 2016-07-04
Changing this.colums to this.columns

sandro-k:
I expect `this.colums` to be a typo and should be `this.columns`. This makes it possible to set the `columns` directly on the Web Component.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/397)

### DIFF
--- a/vaadin-grid.html
+++ b/vaadin-grid.html
@@ -1115,7 +1115,7 @@ See the [demo](demo/index.html) for use case examples.
       if (this.hasAttribute('selection-mode')) {
         this.selection.mode = this.getAttribute('selection-mode');
       }
-      if (this.colums) {
+      if (this.columns) {
         this._grid.setColumns(this.columns);
       } else {
         this.columns = this._grid.getColumns();


manolo:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/397#-:-KLtZgud_si5YqZdGzYj:1417088721)*

tomivirkki:
Hi
We're currently trying to come up with a test case for this (one that would fail with the typo in place) but without success so far. Because of that we're now considering dropping the check altogether (if it's in fact not needed).

In the description you stated that "This makes it possible to set the columns directly on the Web Component". Can you possibly provide example code for this case or verify that your use case works without this check?

sandro-k:
Hi @tomivirkki,
I just checked and Lines 1118-1121 can be removed, just keep 
```
this.columns = this._grid.getColumns();
```

otherwise Polymer starts throwing errors. My setup looks like this: 

```
<vaadin-grid columns=[[gridCol]]>
  <table>
    <tbody>
    <tr>
      <td>Foo</td>
      <td>Bar</td>
    </tr>
    </tbody>
  </table>
</vaadin-grid>

gridCol: {
          type: Array,
          value: function () {
            return [
              {
                name: ''foo",
                renderer: function (cell) {
                  cell.element.innerHTML = '<h1>' + cell.data + '</h1>';
                }
              },
              {
                "name": "Bar"
              }
            ]
          }
        }
```

I figured out that setting it up this way would also works with the old version, because `_columnsChanged` is triggered and sets my columns.

sandro-k:
:+1:

Issue #407, opened 2016-07-11
Capturing click event on a cell


misoucouma:
Hi,
Is there any built in event that we can capture when a user clicks on a cell ?
I am trying to add editing capabilities to the grid, so I need to to capture the click event and the underlying object used in binding.

mshahidkhan:
You can use cellRenderer, append a custom element (button/anchor) in the cell and user your own on-click/on-tap action.
Check this out -
http://codepen.io/mskhan0781/pen/jqQBQr?editors=1111

manolo:
@mshahidkhan :  That is a good example thanks for sharing.
@misoucouma : The grid handles all click events of its children in order to focus and select the component. The way to capture events is that you use cellRenderer as @mshahidkhan suggest, but you have to stop-propagation to avoid the grid reusing the event.
There is a PR #408 in review to ignore clicks over certain html elements, that would mitigate that you have to stop-propagation, but anyway you need to use cellRenderers.
Since the grid reuses cells because of performance you have to be aware that you add the event once.

misoucouma:
that was exact what I was looking for. I actually had a trouble with the event firing more than once.
I solved  the issue by attaching the click event to the lower level child of the TD cell.  Also, to get the column name from the cell that has received a click, cols of the colgroup must have attribute "name"

manolo:
Well, you have the cell info including column name, column index, data, row index, etc, in the moment the current cell is renderer.
Also you can figure-out when you have to setup the dom-tree and start listening for events.

Hope this example helps you:

```
column[0].renderer = function(cell) {
  var first = !cell.element.hasChildNodes();

  if (first) {
    // Add the DOM structure the first time to improve performance.
    cell.element.innerHTML = "<my-element></my-element>";
  }
  var elem = cell.element.querySelector("my-element");
  if (first) {
    // Add event listener first time
    elem.addEventListener('click', function(ev){
      // Avoid grid handling click for selection
      ev.stopPropagation();
      // We have current cell info in the element
      columnName = elem.cell.columnName;
      columnIndex = elem.cell.index;
      rowIndex = elem.row.index;
      myEventHandler(rowIndex .... );
      
    });
  }

  // Save cell info in element because the cell instance is reused.
  elem.cell = {
    index: cell.index,
    columnName: cell.columnName,
    data: cell.data,
    row: {index: cell.row.index, data: cell.row.data}
  };
};

```

jouni:
[...]

Issue #398, opened 2016-07-05
Single selection/row click focus doesn't work in IE11


tomivirkki:

manolo:
Single selection works to me in IE11. It even works when having custom cell renderers.
I have tried the 2nd selection example in cdn changing the attribute `selection-mode` to `single` and I see log in console.
Can you provide a tip to reproduce it?

PR #387, opened 2016-06-13
Prevent focus from being stolen for focusable elements inside cells

Saulis:
Fixes #380

Note for reviewers: didn't add automated tests because of the unreliability of focus on our testing environment.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/387)

PR #408, opened 2016-07-12
Fix/row focus

### LINKED ISSUE #402: Row focus and keyboard select don't work correctly
- A row can be focused by clicking but the focused row doesn't change with up/down arrows
- A focused row doesn't get selected with spacebar

manolo:
Fixes some regressions introduced in #387, fixes #402 #398 #407

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/408)


manolo:
Note for reviewers:
- Testing the current focused element is worst than doing nothing. It does not work in IE, and it has regression problems than prevent the keyboard navigation work
- The most reliable way is do not consider clicks on well-known elements: input, button, checkbox etc. If we do with native elements we do not consider other things like paper-input paper-button so that is the reason of having a regex
  - There are things not covered like a clickable div. In this case the user has the responsability to call the stop-propagation
- Another option is not to do anything, but document that any interactive element put inside the grid should listen to click events and  stop-propagation

- I'll add some tests after the discussion

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-:-KMYGInKT9DDkukvXHrF:-1770720256)*

platosha:
As discussed p2p with @manolo, we should also try these ways:

1. Patch the Grid.java to remove the unnecessary .focus() call on click instead of workarounding it.

2. When something gets focused, it sends a "focusin" event, that propagates up to parents. We can catch it and store the focused element for checking in the click listener. For the Firefox, the "focusin" event does not exist, but we can use "focus".

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-:-KMZ0j73UI1vm-sxm8Ch:-751428573)*

platosha:
Reviewed 1 of 4 files at r4, 1 of 1 files at r6, 1 of 4 files at r7, 1 of 2 files at r9.
Review status: 4 of 6 files reviewed at latest revision, 7 unresolved discussions.

---

*[java/src/main/java/com/vaadin/elements/common/js/JS.java, line 24 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiaXrNOMnZr7d7KNbG:-KMiaXrOJMyX5br-7Od_:-1658350354) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/common/js/JS.java#L24)):*
> ```Java
> 
>     public static final boolean ISIE = GQuery.browser.msie
>             || GQuery.browser.mozilla && Window.Navigator.getUserAgent().toLowerCase().contains("trident");
> ```

Is it also true for Edge? Do we need our workaround to work for Edge?

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 202 \[r8\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiM7gxhM4Ab2HIkvZT:-KMiM7gxhM4Ab2HIkvZU:38298336) ([raw file](https://github.com/vaadin/vaadin-grid/blob/578bbed546bfce278723e4a5ab68f9892970fef2/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L202)):*
> ```Java
>         String fTag = focused.getTagName().toLowerCase();
>         return !focused.hasAttribute("disabled") && (focused.hasAttribute("tabindex")
>                 || fTag.matches("button|input|select|option|textarea|object|iframe|label")
> ```

Labels and options are not focusable. Please exclude from here.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 97 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMia0PA6iBI8qh2yd89:-KMia0PA6iBI8qh2yd8A:-1482966193) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L97)):*
> ```Java
>     }
> 
>     public native void refreshHeader()
> ```

Note that the refreshHeader() and refreshFooter() methods were moved for some reason. Unnecessary change, maybe undo.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 174 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMi_ocVwrvSYBhySwRw:-KMi_ocVwrvSYBhySwRx:-2126390534) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L174)):*
> ```Java
>                         // Forcing focus on the grid will mitigate the issue.
>                         getElement().focus();
>                     } else if (getElement().isOrHasChild(target)) {
> ```

Should be `...isOrHasChild(focused)...`

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 187 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiYETs_rAIEqHd2gNR:-KMiYETs_rAIEqHd2gNS:-2122482895) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L187)):*
> ```Java
> 
>     private void workaroundFlexParentsIE11(Element target) {
>         // In IE11 flex items can also receive focus and spoof the
> ```

“... flex items, as well as table cells, ...”. Table cells should be mentioned in the comment.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 197 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiXRVWoIeRDanKAhke:-KMiXRVWoIeRDanKAhkf:357472369) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L197)):*
> ```Java
>             }
>         });
>         tree.eq(tree.size() - 1).prop("disabled", true).delay(0, lazy().prop("disabled", false).done());
> ```

After timeout, we should reset disabled to the previous state rather than to false.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 202 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiYs_jxNemMn4lLGfs:-KMiYs_kvSYLmwvNt8c1:-1392057240) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L202)):*
> ```Java
>     private boolean isFocusable(Element focused) {
>         String fTag = focused.getTagName().toLowerCase();
>         return !focused.hasAttribute("disabled") && (focused.hasAttribute("tabindex")
> ```

It’s possible to disable an element with a property. We should probably rely on the value of disabled property instead of the attribute presence.

Btw, we can probably remove the disabled check here at all. If something is disabled, it can not be focused natively, even in IE when it's a td or a flex item. So `!!document.activeElement.disabled` should never be true. Our IE workaround uses this feature. In the workaround method you can check `e.disabled` directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408)*

manolo:
Review status: 4 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[java/src/main/java/com/vaadin/elements/common/js/JS.java, line 24 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiaXrNOMnZr7d7KNbG:-KMix9rxGyR7AURoUHS8:1057006177) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/common/js/JS.java#L24)):*
> Is it also true for Edge? Do we need our workaround to work for Edge?
</details>
Edge is not trident.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 202 \[r8\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiM7gxhM4Ab2HIkvZT:-KMix7lVi8xZziihaMfY:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/578bbed546bfce278723e4a5ab68f9892970fef2/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L202)):*
> Labels and options are not focusable. Please exclude from here.
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 97 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMia0PA6iBI8qh2yd89:-KMiv82_6KHKZ_rlXp5b:-352447869) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L97)):*
> Note that the refreshHeader() and refreshFooter() methods were moved for some reason. Unnecessary change, maybe undo.
</details>
It's intentional. It's more readable and error prune to have those hacks by the method they are used.
Also I changed their signature which was wrong.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 174 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMi_ocVwrvSYBhySwRw:-KMivtvGkkLGQwnPbG7K:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L174)):*
> Should be `...isOrHasChild(focused)...`
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 187 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiYETs_rAIEqHd2gNR:-KMiwbpkFT0SJGSlMx52:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L187)):*
> “... flex items, as well as table cells, ...”. Table cells should be mentioned in the comment.
</details>
Done.

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 197 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiXRVWoIeRDanKAhke:-KMiqZIXT6DnolHW4_JJ:308140841) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L197)):*
> After timeout, we should reset disabled to the previous state rather than to false.
</details>
Actually the element has `disabled=true` otherwise it would not be in the list because of the !isFocusable check in the filter predidate

---

*[java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java, line 202 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMiYs_jxNemMn4lLGfs:-KMir-HxhTLH500raIEq:777578834) ([raw file](https://github.com/vaadin/vaadin-grid/blob/22014ab857b436e2068418332d1d1d8c7c21d167/java/src/main/java/com/vaadin/elements/grid/ViolatedGrid.java#L202)):*
> It’s possible to disable an element with a property. We should probably rely on the value of disabled property instead of the attribute presence.
> 
> Btw, we can probably remove the disabled check here at all. If something is disabled, it can not be focused natively, even in IE when it's a td or a flex item. So `!!document.activeElement.disabled` should never be true. Our IE workaround uses this feature. In the workaround method you can check `e.disabled` directly.
</details>
we need this check here for the predicate, but yes it's better to check the property instead

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408)*

platosha:
:lgtm: We should probably squash the changes before merging.

---

Reviewed 1 of 4 files at r7, 3 of 3 files at r13.

---

*[test/grid-rendering-light-dom.html, line 370 \[r13\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMwpx9aQ0M_SZcSbbqq:-KMwpx9aQ0M_SZcSbbqr:1702488435) ([raw file](https://github.com/vaadin/vaadin-grid/blob/d27bce69b5fd19c6308b6a201396bc48ffdae0b2/test/grid-rendering-light-dom.html#L370)):*
> ```HTML
>           fullClick(elem);
>           waitUntil(function() {
>             // IE11 would return the first flex-element or table cell
> ```

“IE11 would” is confusing here, though the conditions are explained in the test below. I’d mention the workaround here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-:-KMwrB_jTIFVO4FpH0XG:-1074700192)*

manolo:
Review status: 3 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[test/grid-rendering-light-dom.html, line 370 \[r13\]](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408#-KMwpx9aQ0M_SZcSbbqq:-KN0h1bM8HwP7_Q0KApn:-499146009) ([raw file](https://github.com/vaadin/vaadin-grid/blob/d27bce69b5fd19c6308b6a201396bc48ffdae0b2/test/grid-rendering-light-dom.html#L370)):*
> “IE11 would” is confusing here, though the conditions are explained in the test below. I’d mention the workaround here.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408)*

manolo:
Reviewed 1 of 1 files at r6, 2 of 4 files at r7, 3 of 3 files at r15.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/408)*

PR #423, opened 2016-08-17
Sort column in-place when using items array

### LINKED ISSUE #212: Grid should offer default sorting if using array data source
It would be very convenient if the Grid supplied a default comparator for sorting Strings/Numbers/Dates in natural order. The user could always supply their own version if they wanted, but it would make the 90%+ use case much easier.

Saulis:
Fixes #212

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/423)

### DIFF
--- a/demo/sort.html
+++ b/demo/sort.html
@@ -72,8 +72,8 @@ layout: page
   <section>
     <h3>Sorting the Data Source</h3>
     <p>As mentioned in the previous example, there are multiple ways to change <code>vaadin-grid</code> sort order.
-    None of them will, however, automatically order any data inside <code>vaadin-grid</code>. The data source implementation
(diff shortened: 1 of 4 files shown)


tehapo:
Added my small refactoring commit after first discussing the changes with @saulis. Otherwise this looks good to me.

Maybe the side-effect of sorting the original array might be emphasised a bit more in the documentation, but I'm ok with this version.

PS. Nice touch on using `reduce()` to support multiple sorted columns. :+1:

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/423#-:-KPRi8_S6H8VJGPu8gzL:b4r3wv7)*

PR #469, opened 2016-09-19
Add methods to get row/col indexes given a child element

manolo:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/469)


tomivirkki:
If you need the row/column indexes of a cell element, rather include them as data-attributes on the cell element itself. Having a public API in `vaadin-grid` that accepts cell elements (that are supposed to be hidden in the local DOM) as parameters doesn't sound right.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-:-KS14-2fkpN077qE6E71:b-avnj6w)*

manolo:
Why to hit performance by adding attributes to all cells each time they are rendered or moved in the grid? most use cases of the the grid would not use those data attributes.

Doesn't it make more sense that given and event, the user can ask for the grid where that event was originated? 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-:-KS1IIqtbT_pTJHIywsS:b-onok32)*

tomivirkki:
If this feature is intended for vaadin-context-menu's purposes only, then vaadin-grid should fire a context-aware event instead.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-:-KS1QUK--DtCPxN-dBPx:bbaepon)*

tomivirkki:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid.html, line 1132 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6VtAVynrNG_V0_Gpc:-KS6VtAVynrNG_V0_Gpd:b-jcs0ac) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1132)):*
> ```HTML
>       detailedEvents: {
>         type: Array,
>         default: []
> ```

should say
`value: function() {
  return [];
}`

---

*[vaadin-grid.html, line 1136 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6X6F1CP8xT95dGjo3:-KS6X6F1CP8xT95dGjo4:b-jcs0ac) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1136)):*
> ```HTML
> 
>       _currentDetailed: {
>         value: []
> ```

should say
`value: function() {
  return [];
}`

---

*[vaadin-grid.html, line 1184 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WEEvoRyTlCGpWvqg:-KS6WEEvoRyTlCGpWvqh:b-2sz7qk) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1184)):*
> ```HTML
>        * @event detailed-event
>        * @param {Object} detail
>        *  @param {String} detail.event The original event.
> ```

Why do we want to include the original event (and the reference to the internal element at the same time), Type is Wrong

---

*[vaadin-grid.html, line 1187 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WgXED3IAqFEqzDyP:-KS6WgXED3IAqFEqzDyQ:bq0hn6q) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1187)):*
> ```HTML
>        *  @param {Number} detail.column The column index where the element is located.
>        *  @param {Number} detail.row The row index where the element is located
>        *  @param {Number} detail.section The section name in the grid (header|body|footer)
> ```

This rather seems to be a String

---

*[vaadin-grid.html, line 1188 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6Wk6DBykq95TLcV65:-KS6Wk6DBykq95TLcV66:b9fspqx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1188)):*
> ```HTML
>        *  @param {Number} detail.row The row index where the element is located
>        *  @param {Number} detail.section The section name in the grid (header|body|footer)
>        *  @param {Number} detail.data The row data if the event was on the body, otherwise undefined
> ```

Data isn't usually a number

---

*[vaadin-grid.html, line 1478 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WpDtKJUxtNU7RYpB:-KS6WpDtKJUxtNU7RYpC:bl8wkn3) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1478)):*
> ```HTML
>     _detailedChanged: function() {
>       // childElementCount is 0 in non-shadow
>       var node = this.root.childElementCount ? this.root : this;
> ```

The grid element is always there so we can listen to that
`var node = this._grid.getGridElement();`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid.html, line 1184 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WEEvoRyTlCGpWvqg:-KS6XVTzqIQCP2s-eDsH:ba9elqk) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1184)):*
> Why do we want to include the original event (and the reference to the internal element at the same time), Type is Wrong
</details>
Yes, type is wrong. The element is in the event. The event is include because the user would want to stopPropagationg or preventDefault.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

tomivirkki:
*[vaadin-grid.html, line 1184 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WEEvoRyTlCGpWvqg:-KS6Y7YhULUhJgmI2o0q:b-ium4cx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1184)):*
> Yes, type is wrong. The element is in the event. The event is include because the user would want to stopPropagationg or preventDefault.
</details>
Right, would it be better that we always stopped + canceled it ourselves (internally) for detailed events

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid.html, line 1132 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6VtAVynrNG_V0_Gpc:-KS6ZCSN5LIddeySOt-n:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1132)):*
> should say
> `value: function() {
>   return [];
> }`
</details>
Done.

---

*[vaadin-grid.html, line 1136 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6X6F1CP8xT95dGjo3:-KS6ZCpVNl7VeNSbnt_4:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1136)):*
> should say
> `value: function() {
>   return [];
> }`
</details>
Done.

---

*[vaadin-grid.html, line 1184 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WEEvoRyTlCGpWvqg:-KS6ZHOJ-fBRX0yaqfoD:bjjkywn) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1184)):*
> Right, would it be better that we always stopped + canceled it ourselves (internally) for detailed events
</details>
I see cases where it would be nice, and other where they  are not. 
For instance, if you want to do something when the grid is clicked, but you want the grid to continue  handling the selection event, etc.

---

*[vaadin-grid.html, line 1187 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WgXED3IAqFEqzDyP:-KS6ZZSu9HQfZx9WdpY9:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1187)):*
> This rather seems to be a String
</details>
Done.

---

*[vaadin-grid.html, line 1188 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6Wk6DBykq95TLcV65:-KS6Zo5ApFLNfSls7MAa:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1188)):*
> Data isn't usually a number
</details>
Done.

---

*[vaadin-grid.html, line 1478 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WpDtKJUxtNU7RYpB:-KS6ZqFYl9tYUo9yVGLa:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1478)):*
> The grid element is always there so we can listen to that
> `var node = this._grid.getGridElement();`
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

tomivirkki:
*[vaadin-grid.html, line 1184 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WEEvoRyTlCGpWvqg:-KS6ajab08tLXh-HUqUy:b-orumjy) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1184)):*
> I see cases where it would be nice, and other where they  are not. 
> For instance, if you want to do something when the grid is clicked, but you want the grid to continue  handling the selection event, etc.
</details>
Good point

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid.html, line 1184 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS6WEEvoRyTlCGpWvqg:-KS6cK78h3nQw5kfEJfW:b-6mggu8) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3ef66d5b744b4deb1d7f4d170790410fadda5e05/vaadin-grid.html#L1184)):*
> Good point
</details>
We could selectively prevent events but some events that already uses the like click, touch, scroll. But I'd rather pass that responsibility to the developer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid.html, line 1484 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSAOUuN4XN646ikq7YA:-KSAOUuN4XN646ikq7YB:bfwayu) ([raw file](https://github.com/vaadin/vaadin-grid/blob/04986b59c4263f0f3120b9b39067f1e823a62d47/vaadin-grid.html#L1484)):*
> ```HTML
>       for (var i = this._currentDetailed.length; i > 0; i--) {
>         var name = this._currentDetailed[i - 1];
>         if (!this.detailedEvents.includes(name)) {
> ```

@tehapo I guess this is another case that will fail in IE11, right ? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

tehapo:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid.html, line 1484 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSAOUuN4XN646ikq7YA:-KSAPQsPEnHg9kUuJ20Z:b-gdmped) ([raw file](https://github.com/vaadin/vaadin-grid/blob/04986b59c4263f0f3120b9b39067f1e823a62d47/vaadin-grid.html#L1484)):*
> @tehapo I guess this is another case that will fail in IE11, right ? 
</details>
That's right, `includes` method on arrays is part of ES2016 spec and while already supported on most browsers, it is not supported in IE11.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

tehapo:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-grid.html, line 1186 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSAQluATffzZVNxJy-m:-KSAQluATffzZVNxJy-n:bjwjzwx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/04986b59c4263f0f3120b9b39067f1e823a62d47/vaadin-grid.html#L1186)):*
> ```HTML
>        * Fired when an event registered in the `detailedEvents` array is fired.
>        *
>        * @event detailed-event
> ```

Not sure if I like the event name.

Also, should it be prefixed with `grid-` or `vaadin-grid-`? I have no good suggestion for the name in mind, but just thinking out loud here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-grid.html, line 1186 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSAQluATffzZVNxJy-m:-KSAexpadGzKCB9l3pFW:bpmauq9) ([raw file](https://github.com/vaadin/vaadin-grid/blob/04986b59c4263f0f3120b9b39067f1e823a62d47/vaadin-grid.html#L1186)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> Not sure if I like the event name.
> 
> Also, should it be prefixed with `grid-` or `vaadin-grid-`? I have no good suggestion for the name in mind, but just thinking out loud here.
</details>
Yeah, I didn't take this in account, but it makes sense that we could have nested elements in a UI and they all sends up event-details. I don't know whether it it is a real or edge use-case, though

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-grid.html, line 1484 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSAOUuN4XN646ikq7YA:-KSAfTmA-0NIU3dfoMGD:brhkprj) ([raw file](https://github.com/vaadin/vaadin-grid/blob/04986b59c4263f0f3120b9b39067f1e823a62d47/vaadin-grid.html#L1484)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> That's right, `includes` method on arrays is part of ES2016 spec and while already supported on most browsers, it is not supported in IE11.
</details>
Seems like indexOf is, so changing to it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
> @manolo I think the PR is incomplete. As discussed today, I wish:
> 1) original names prefixed: detailed-click... rather then always the same detailed-event.
> 2) preventing the detailed event should prevent the original one.

@platosha Done.

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-:-KSFXw4CxmSGY2ANEoWG:bnbk5if)*

platosha:
Reviewed 2 of 6 files at r2, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-grid.html, line 1480 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSF_mXzim_Oma9x_vqr:-KSF_mXzim_Oma9x_vqs:bwp8llv) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1480)):*
> ```HTML
> 
>     _onDetailed: function(ev) {
>       var evt = this.fire('detailed-' + ev.type, this._grid.getEventDetails(ev));
> ```

Currently the names `ev` and `evt` don’t really tell any difference between them. Maybe rename `evt` to `detailedEvent`?

---

*[vaadin-grid.html, line 1482 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSF_YZrRSFTdERzTFh4:-KSF_YZsan1rrIYZiTjf:bkocz5p) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1482)):*
> ```HTML
>       var evt = this.fire('detailed-' + ev.type, this._grid.getEventDetails(ev));
>       if (evt.defaultPrevented) {
>         ev.stopPropagation();
> ```

Not sure about `stopPropagation()`. Is it needed? What is the use case?

---

*[vaadin-grid.html, line 1490 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFb5y8VViRUHgvO7PS:-KSFb5y8VViRUHgvO7PT:b-rv366c) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1490)):*
> ```HTML
>       var node = this._grid.getGridElement();
>       for (var i = this._currentDetailed.length; i > 0; i--) {
>         var name = this._currentDetailed[i - 1];
> ```

Walking arrays backwards looks somewhat disturbing, as well as using the expresson `[i - 1]` as the index. Although the code understandable for me, it is harder to read. Perhaps we should’t do that if there is no need.

As I understood, backwards walk is chosen here for simplicity of item removal. If so, good, but still would be better to have the plain index `[i]` of the current item instead of the expression `[i - 1]`;

---

*[vaadin-grid.html, line 1493 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFbQuKJBcciYDKzi4A:-KSFbQuKJBcciYDKzi4B:b-vb08z) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1493)):*
> ```HTML
>         if (this.detailedEvents.indexOf(name) < 0) {
>           this.unlisten(node, name, '_onDetailed');
>           this._currentDetailed.pop();
> ```

Well, well, well... Looks like a mistake for me. Consider:

```javascript
// initial value
grid.detailedEvents = ['click', 'contextmenu'];
// _currentDetailed value becomes: ['click', 'contextmenu'], OK

// then someone removes non-last event from the array
grid.detailedEvents = ['contextmenu'];
// _currentDetailed value becomes: ['click'], wrong!
```

On the second step of this loop, it will be detected that the `click` event was removed, so it should remove `click` from the `_currentDetailed`. But the removal line calls `pop()`, which doesn’t receive the array index and just removes the last item, which is `contextmenu`.

Here should be `splice` removal for the current item index instead of `pop`, I think.

---

*[vaadin-grid.html, line 1501 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFlQCs0sH7Bfkjkwq2:-KSFlQCs0sH7Bfkjkwq3:b-31e0mc) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1501)):*
> ```HTML
>           this._currentDetailed.push(name);
>         }
>       }.bind(this));
> ```

I would suggest to do it this way for better readibility: http://jsbin.com/tocedilape/edit?js,console

---

*[test/grid-properties.html, line 317 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFm51nYhicgwF9MOdS:-KSFm51ordWAy2RSnUu5:b-aczl05) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/test/grid-properties.html#L317)):*
> ```HTML
>       MockInteractions.tap(th);
>     });
> ```

Should we also test an array of multiple detailed events, just to make sure that all of them are fired with details?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

Saulis:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[vaadin-grid.html, line 1123 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSG-k9GtS1ayr9qCcia:-KSG-k9GtS1ayr9qCcib:bv4u3sg) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1123)):*
> ```HTML
>       },
> 
>       _currentDetailed: {
> ```

This shouldn't be needed if `detailedEvents` is properly observed.

---

*[vaadin-grid.html, line 1487 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFzyf_AiXkrn78XXcN:-KSFzyf_AiXkrn78XXcO:b7va8tn) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1487)):*
> ```HTML
>     },
> 
>     _detailedChanged: function() {
> ```

Please make this into a proper splice observer, which checks the change event for removed and added items. Here's a good example: https://www.polymer-project.org/1.0/docs/devguide/observers#array-observation. 

Also, I suggest having a separate observer method for situations where the whole array is switched.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[vaadin-grid.html, line 1123 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSG-k9GtS1ayr9qCcia:-KSG0IZjhO8qxZKgjwgi:b-yyoutx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1123)):*
> This shouldn't be needed if `detailedEvents` is properly observed.
</details>
what do you mean? we need a way to figure out what events we were previously listening to.
We can use the polymer lowlevel  changes api but it's more complex code.

---

*[vaadin-grid.html, line 1480 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSF_mXzim_Oma9x_vqr:-KSG0j0IvPOwh8lTq1NY:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1480)):*
> Currently the names `ev` and `evt` don’t really tell any difference between them. Maybe rename `evt` to `detailedEvent`?
</details>
Done.

---

*[vaadin-grid.html, line 1482 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSF_YZrRSFTdERzTFh4:-KSFoJt9n-E0V_xCvTwI:b-fy6bfu) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1482)):*
> Not sure about `stopPropagation()`. Is it needed? What is the use case?
</details>
For instance  `detailedEvents=['click'];` and the user wants to avoid that the grid ignore that click (selection in body, sort in headers) 

---

*[vaadin-grid.html, line 1487 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFzyf_AiXkrn78XXcN:-KSG0zUTWV64jfVSitrk:b-dgz69q) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1487)):*
> Please make this into a proper splice observer, which checks the change event for removed and added items. Here's a good example: https://www.polymer-project.org/1.0/docs/devguide/observers#array-observation. 
> 
> Also, I suggest having a separate observer method for situations where the whole array is switched.
</details>
As I said the changes api introduces more complexity to our code.

---

*[vaadin-grid.html, line 1490 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFb5y8VViRUHgvO7PS:-KSFrJmq7IY98P0e43_9:b774gyy) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1490)):*
> Walking arrays backwards looks somewhat disturbing, as well as using the expresson `[i - 1]` as the index. Although the code understandable for me, it is harder to read. Perhaps we should’t do that if there is no need.
> 
> As I understood, backwards walk is chosen here for simplicity of item removal. If so, good, but still would be better to have the plain index `[i]` of the current item instead of the expression `[i - 1]`;
</details>
Walking arrays is a very common code that all developers must be used with. Using `i` index instead of `i - 1`

---

*[vaadin-grid.html, line 1493 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFbQuKJBcciYDKzi4A:-KSFrKfHYIQXC-PNvSiy:b1xho2a) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1493)):*
> Well, well, well... Looks like a mistake for me. Consider:
> 
> ```javascript
> // initial value
> grid.detailedEvents = ['click', 'contextmenu'];
> // _currentDetailed value becomes: ['click', 'contextmenu'], OK
> 
> // then someone removes non-last event from the array
> grid.detailedEvents = ['contextmenu'];
> // _currentDetailed value becomes: ['click'], wrong!
> ```
> 
> On the second step of this loop, it will be detected that the `click` event was removed, so it should remove `click` from the `_currentDetailed`. But the removal line calls `pop()`, which doesn’t receive the array index and just removes the last item, which is `contextmenu`.
> 
> Here should be `splice` removal for the current item index instead of `pop`, I think.
</details>
good catch ! Using splice instead.

---

*[vaadin-grid.html, line 1501 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFlQCs0sH7Bfkjkwq2:-KSG1OFRxLwV3B2RVd1b:bd65rzs) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1501)):*
> I would suggest to do it this way for better readibility: http://jsbin.com/tocedilape/edit?js,console
</details>
Simplified my code to throw out the old storage as you suggested, but using slice to clone instead of reduce.

---

*[test/grid-properties.html, line 317 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFm51nYhicgwF9MOdS:-KSFzfbY-xYbKAjb2Pwt:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/test/grid-properties.html#L317)):*
> Should we also test an array of multiple detailed events, just to make sure that all of them are fired with details?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

platosha:
:lgtm:

---

Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-:-KSH79LQxmTq1V87b7JH:bnfp4nl)*

Saulis:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-grid.html, line 1487 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KSFzyf_AiXkrn78XXcN:-KSLizmZNduOFpPr67Ac:b-de6vu4) ([raw file](https://github.com/vaadin/vaadin-grid/blob/92d284da6996dd219f92ecf5c6fb40ec66ec83bc/vaadin-grid.html#L1487)):*
> As I said the changes api introduces more complexity to our code.
</details>
The revision 5 looks better, I'm OK with that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

tomivirkki:
*[vaadin-grid.html, line 1488 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS_QxWbV1FDkJmfWmUd:-KS_QxWbV1FDkJmfWmUe:b-uzj5h8) ([raw file](https://github.com/vaadin/vaadin-grid/blob/01527eccc366da1a3a86f996dd251de46b278f25/vaadin-grid.html#L1488)):*
> ```HTML
> 
>     _detailedChanged: function() {
>       var node = this._grid.getGridElement();
> ```

So what I suggested in the earlier discussion was to listen to the vaadin-grid itself for the events. Basically replacing the 'node' variable with 'this' and then fetching the original target with Polymer API's in _onDetailed function.

This way the 'detailed-' event will have a safe reference to the "vaadin-grid level" event that's already retargeted and safe to pass onwards.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-grid.html, line 1488 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS_QxWbV1FDkJmfWmUd:-KS_UfHSJC0qul7K8PE1:b-lcrr1a) ([raw file](https://github.com/vaadin/vaadin-grid/blob/01527eccc366da1a3a86f996dd251de46b278f25/vaadin-grid.html#L1488)):*
> So what I suggested in the earlier discussion was to listen to the vaadin-grid itself for the events. Basically replacing the 'node' variable with 'this' and then fetching the original target with Polymer API's in _onDetailed function.
> 
> This way the 'detailed-' event will have a safe reference to the "vaadin-grid level" event that's already retargeted and safe to pass onwards.
</details>
I don't follow, how do I know the exact element that received the event in the shadow dom? I need it for getting the grid.row/column/data 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

platosha:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-grid.html, line 1488 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS_QxWbV1FDkJmfWmUd:-KS_Uo_R7tEeyPnVRHzo:b-8ymbdy) ([raw file](https://github.com/vaadin/vaadin-grid/blob/01527eccc366da1a3a86f996dd251de46b278f25/vaadin-grid.html#L1488)):*
> I don't follow, how do I know the exact element that received the event in the shadow dom? I need it for getting the grid.row/column/data 
</details>
@manolo try `Polymer.dom(event).rootTarget`
see: https://www.polymer-project.org/1.0/docs/devguide/events#retargeting

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

platosha:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-grid.html, line 1488 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS_QxWbV1FDkJmfWmUd:-KSaXbxKgJmu1AMDHJF2:b32w0x) ([raw file](https://github.com/vaadin/vaadin-grid/blob/01527eccc366da1a3a86f996dd251de46b278f25/vaadin-grid.html#L1488)):*
> @manolo try `Polymer.dom(event).rootTarget`
> see: https://www.polymer-project.org/1.0/docs/devguide/events#retargeting
</details>
OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

manolo:
Review status: 1 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-grid.html, line 1488 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469#-KS_QxWbV1FDkJmfWmUd:-KSabW9gfghuUtcy1Tb_:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/01527eccc366da1a3a86f996dd251de46b278f25/vaadin-grid.html#L1488)):*
> OK
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/469)*

Issue #441, opened 2016-08-25
Refactor colspan API to <vaadin-column-group>


jouni:
``` html
<vaadin-grid>

  <vaadin-grid-column frozen="start">
    <template class=“header>Index</template>
    <template>[[index]]</template>
  </vaadin-grid-column>

  <vaadin-grid-column-group frozen="start">
    <template class=“header>Name</template>

    <vaadin-grid-column>
      <template class=“header>First Name</template>
      <template>[[item.firstName]]</template>
    </vaadin-grid-column>
    <vaadin-grid-column>
      <template class=“header>Last Name</template>
      <template>[[item.lastName]]</template>
    </vaadin-grid-column>
    <vaadin-grid-column>
      <template class=“header>Nickname</template>
      <template>[[item.nickName]]</template>
    </vaadin-grid-column>
  </vaadin-grid-column-group>
  ...
</vaadin-grid>
```
#### `frozen` attribute/property

The `frozen` attribute is only allowed on the top-most level (direct child of `<vaadin-grid>`). All others are ignored. This seems like the approach that causes the least amount of confusion for users.

The attribute can have `start` and `end` values, to specify where to freeze the columns. The default should be `start` (if unspecified).

jouni:
Run a quick hallway test for the proposed API to see if people understand what is happening.

jouni:
Added a section about frozen columns in the description.

Saulis:
@jouni we were thinking about not including the `start` and `end` parameters at this point as the right aligned frozen columns isn't finalized yet.?

jouni:
@Saulis, sure. I was about to write that in the description, but thought you’d do it like that anyway, so I didn’t bother 😄

Saulis:
Got some comments about the API from Artur:

1) How do you know/define on which header row "index" header in this case goes? 1st or 2nd?

2) If you want to group footers differently from headers, how does that happen?

3) In case you want to dynamically group/ungroup columns, you kinda need to rebuild the whole columns configuration by wrapping/unwrapping some columns.

In general he found the column API a bit confusing, a plain `<table>` would've made more sense for him.

jouni:
Hmm, good comments.

1) My idea was that they always start to stack from bottom up. So if you have a plain column, without any grouping, that header is closest to the data rows. From there on, each new column group would be added on top of the previous groups header. If a group doesn’t specify a header, then that header cell is empty.

[...]

PR #471, opened 2016-09-28
Optimize table-row _updateCells 

manolo:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/471)

### DIFF
--- a/test/column.html
+++ b/test/column.html
@@ -99,39 +99,32 @@
             expect(container.flex).to.eql(2);
           });
 
-          it('should be bound to header cells', function(done) {
+          it('should be bound to header cells', function() {
             column.flex = 2;
+
             var header = grid.$.scroller.$.header;
-            grid.async(function(){
-              expect(header._rows[0].cells[0].style.flexGrow).to.eql('2');
-              expect(header._rows[1].cells[0].style.flexGrow).to.eql('3'); // colspan 2 + 1
-              done();
-            });
+
+            expect(header._rows[0].cells[0].style.flexGrow).to.eql('2');
+            expect(header._rows[1].cells[0].style.flexGrow).to.eql('3'); // colspan 2 + 1
           });
 
-          it('should be bound to row cells', function(done) {
+          it('should be bound to row cells', function() {
             column.flex = 2;
-            grid.async(function() {
-              expect(grid.$.scroller._physicalItems[0].cells[0].style.flexGrow).to.eql('2');
-              done();
-            });
+
+            expect(grid.$.scroller._physicalItems[0].cells[0].style.flexGrow).to.eql('2');
           });
 
-          it('should be bound to row cells on multiSelection', function(done) {
+          it('should be bound to row cells on multiSelection', function() {
             grid.multiSelection = true;
             column.flex = 2;
-            grid.async(function() {
-              expect(grid.$.scroller._physicalItems[0].cells[1].style.flexGrow).to.eql('2');
-              done();
-            });
+
+            expect(grid.$.scroller._physicalItems[0].cells[1].style.flexGrow).to.eql('2');
           });
 
-          it('should be bound to footer cells', function(done) {
+          it('should be bound to footer cells', function() {
             column.flex = 2;
-            grid.async(function() {
-              expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexGrow).to.eql('2');
-              done();
-            });
+
+            expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexGrow).to.eql('2');
           });
         });
 
@@ -140,46 +133,37 @@
             expect(column.width).to.eql('100px');
           });
 
-          it('should notify changes to width', function(done) {
+          it('should notify changes to width', function() {
             column.width = '200px';
-            grid.async(function() {
-              expect(container.width).to.eql('200px');
-              done();
-            });
+
+            expect(container.width).to.eql('200px');
           });
 
-          it('should be bound to header cells', function(done) {
+          it('should be bound to header cells', function() {
             column.width = '200px';
+
             var header = grid.$.scroller.$.header;
-            grid.async(function() {
-              expect(header._rows[0].cells[0].style.flexBasis).to.eql('calc(200px)');
-              done();
-            });
+
+            expect(header._rows[0].cells[0].style.flexBasis).to.eql('calc(200px)');
           });
 
-          it('should be bound to row cells', function(done) {
+          it('should be bound to row cells', function() {
             column.width = '200px';
-            grid.async(function() {
-              expect(grid.$.scroller._physicalItems[0].cells[0].style.flexBasis).to.eql('calc(200px)');
-              done();
-            });
+
+            expect(grid.$.scroller._physicalItems[0].cells[0].style.flexBasis).to.eql('calc(200px)');
           });
 
-          it('should be bound to row cells on multiSelection', function(done) {
+          it('should be bound to row cells on multiSelection', function() {
             grid.multiSelection = true;
             column.width = '200px';
-            grid.async(function() {
-              expect(grid.$.scroller._physicalItems[0].cells[1].style.flexBasis).to.eql('calc(200px)');
-              done();
-            });
+
+            expect(grid.$.scroller._physicalItems[0].cells[1].style.flexBasis).to.eql('calc(200px)');
           });
 
-          it('should be bound to footer cells', function(done) {
+          it('should be bound to footer cells', function() {
             column.width = '200px';
-            grid.async(function() {
-              expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexBasis).to.eql('calc(200px)');
-              done();
-            });
+
+            expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexBasis).to.eql('calc(200px)');
           });
         });
       });
--- a/test/column.html
+++ b/test/column.html
@@ -99,32 +99,39 @@
             expect(container.flex).to.eql(2);
           });
 
-          it('should be bound to header cells', function() {
+          it('should be bound to header cells', function(done) {
             column.flex = 2;
-
             var header = grid.$.scroller.$.header;
-
-            expect(header._rows[0].cells[0].style.flexGrow).to.eql('2');
-            expect(header._rows[1].cells[0].style.flexGrow).to.eql('3'); // colspan 2 + 1
+            grid.async(function(){
+              expect(header._rows[0].cells[0].style.flexGrow).to.eql('2');
+              expect(header._rows[1].cells[0].style.flexGrow).to.eql('3'); // colspan 2 + 1
+              done();
+            });
           });
 
-          it('should be bound to row cells', function() {
+          it('should be bound to row cells', function(done) {
             column.flex = 2;
-
-            expect(grid.$.scroller._physicalItems[0].cells[0].style.flexGrow).to.eql('2');
+            grid.async(function() {
+              expect(grid.$.scroller._physicalItems[0].cells[0].style.flexGrow).to.eql('2');
+              done();
+            });
           });
 
-          it('should be bound to row cells on multiSelection', function() {
+          it('should be bound to row cells on multiSelection', function(done) {
             grid.multiSelection = true;
             column.flex = 2;
-
-            expect(grid.$.scroller._physicalItems[0].cells[1].style.flexGrow).to.eql('2');
+            grid.async(function() {
+              expect(grid.$.scroller._physicalItems[0].cells[1].style.flexGrow).to.eql('2');
+              done();
+            });
           });
 
-          it('should be bound to footer cells', function() {
+          it('should be bound to footer cells', function(done) {
             column.flex = 2;
-
-            expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexGrow).to.eql('2');
+            grid.async(function() {
+              expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexGrow).to.eql('2');
+              done();
+            });
           });
         });
 
@@ -133,37 +140,46 @@
             expect(column.width).to.eql('100px');
           });
 
-          it('should notify changes to width', function() {
+          it('should notify changes to width', function(done) {
             column.width = '200px';
-
-            expect(container.width).to.eql('200px');
+            grid.async(function() {
+              expect(container.width).to.eql('200px');
+              done();
+            });
           });
 
-          it('should be bound to header cells', function() {
+          it('should be bound to header cells', function(done) {
             column.width = '200px';
-
             var header = grid.$.scroller.$.header;
-
-            expect(header._rows[0].cells[0].style.flexBasis).to.eql('200px');
+            grid.async(function() {
+              expect(header._rows[0].cells[0].style.flexBasis).to.eql('calc(200px)');
+              done();
+            });
           });
 
-          it('should be bound to row cells', function() {
+          it('should be bound to row cells', function(done) {
             column.width = '200px';
-
-            expect(grid.$.scroller._physicalItems[0].cells[0].style.flexBasis).to.eql('200px');
+            grid.async(function() {
+              expect(grid.$.scroller._physicalItems[0].cells[0].style.flexBasis).to.eql('calc(200px)');
+              done();
+            });
           });
 
-          it('should be bound to row cells on multiSelection', function() {
+          it('should be bound to row cells on multiSelection', function(done) {
             grid.multiSelection = true;
             column.width = '200px';
-
-            expect(grid.$.scroller._physicalItems[0].cells[1].style.flexBasis).to.eql('200px');
+            grid.async(function() {
+              expect(grid.$.scroller._physicalItems[0].cells[1].style.flexBasis).to.eql('calc(200px)');
+              done();
+            });
           });
 
-          it('should be bound to footer cells', function() {
+          it('should be bound to footer cells', function(done) {
             column.width = '200px';
-
-            expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexBasis).to.eql('200px');
+            grid.async(function() {
+              expect(grid.$.scroller.$.footer._rows[0].cells[0].style.flexBasis).to.eql('calc(200px)');
+              done();
+            });
           });
         });
       });
--- a/test/headers-footers.html
+++ b/test/headers-footers.html
@@ -64,11 +64,14 @@
         expect(getRowCells(headerRows[1])[0].colspan).to.equal(2);
       });
 
-      it('should render colspan correctly', function() {
+      it('should render colspan correctly', function(done) {
         window.getComputedStyle(getRowCells(headerRows[1])[1]).display === 'none';
         var firstRowCells = getRowCells(headerRows[0]);
         var combinedWidth = firstRowCells[0].clientWidth + firstRowCells[1].clientWidth;
-        expect(getRowCells(headerRows[1])[0].clientWidth).to.equal(combinedWidth);
+        grid.async(function() {
+          expect(getRowCells(headerRows[1])[0].clientWidth).to.equal(combinedWidth);
+          done();
+        });
       });
 
       it('should have last row fully visible when scrolled to end', function(done) {
@@ -80,8 +83,6 @@
           done();
         });
       });
-
-
     });
   </script>
 
--- a/test/headers-footers.html
+++ b/test/headers-footers.html
@@ -64,14 +64,11 @@
         expect(getRowCells(headerRows[1])[0].colspan).to.equal(2);
       });
 
-      it('should render colspan correctly', function(done) {
+      it('should render colspan correctly', function() {
         window.getComputedStyle(getRowCells(headerRows[1])[1]).display === 'none';
         var firstRowCells = getRowCells(headerRows[0]);
         var combinedWidth = firstRowCells[0].clientWidth + firstRowCells[1].clientWidth;
-        grid.async(function() {
-          expect(getRowCells(headerRows[1])[0].clientWidth).to.equal(combinedWidth);
-          done();
-        });
+        expect(getRowCells(headerRows[1])[0].clientWidth).to.equal(combinedWidth);
       });
 
       it('should have last row fully visible when scrolled to end', function(done) {
@@ -83,6 +80,8 @@
           done();
         });
       });
+
+
     });
   </script>
 
--- a/vaadin-grid-table-row.html
+++ b/vaadin-grid-table-row.html
@@ -99,33 +99,31 @@
       },
 
       _updateCells: function() {
-        var spanning = 0;
-        this.cells.forEach(function(cell, index) {
-
-          // TODO: reset cell styles to make sure colspan have be changed dynamically.
-          cell.style.display = '';
-          cell.style.flexBasis = this.columns[index].width;
-          cell.style.flexGrow = this.columns[index].flex;
-
-          if (cell.colspan > 1) {
-            spanning = cell.colspan;
-
-            var width = cell.style.flexBasis;
-            var flex = parseInt(cell.style.flexGrow);
-            for (var i = 1; i < spanning; i++) {
-              width += ' + ' + this.columns[index + i].width;
-              flex += this.columns[index + i].flex;
+        // debouncing reduces the call
+        // - from 60 to 4 in a grid with 2 headers with 5 columns.
+        // - from 416 to 30 in the same grid with data and one footer.
+        this.debounce('_updateCells', function() {
+          // columns > cells means that there is an extra col for multiselection
+          var multiSel = this.columns.length > this.cells.length ? 1 : 0;
+          var spanning = 0;
+          this.cells.forEach(function(cell, index) {
+            var width = this.columns[index + multiSel].width;
+            var flex = this.columns[index + multiSel].flex;
+            // spanning > 0 means that this row is already spanned
+            if (!spanning && cell.colspan > 1) {
+              spanning = cell.colspan;
+              for (var i = 1; i < spanning; i++) {
+                width += ' + ' + this.columns[index + multiSel + i].width;
+                flex += this.columns[index + multiSel + i].flex;
+              }
             }
 
             cell.style.flexBasis = 'calc(' + width + ')';
             cell.style.flexGrow = flex;
-          }
+            cell.style.display = (!cell.colspan && spanning > 0) ? 'none' : '';
+            spanning = Math.max(0, spanning - 1);
 
-          if (!cell.colspan && spanning > 0) {
-            cell.style.display = 'none';
-          }
-
-          spanning = Math.max(0, spanning - 1);
+          }.bind(this));
         }.bind(this));
       },
 
--- a/vaadin-grid-table-row.html
+++ b/vaadin-grid-table-row.html
@@ -99,31 +99,27 @@
       },
 
       _updateCells: function() {
-        // debouncing reduces the call
-        // - from 60 to 4 in a grid with 2 headers with 5 columns.
-        // - from 416 to 30 in the same grid with data and one footer.
-        this.debounce('_updateCells', function() {
-          // columns > cells means that there is an extra col for multiselection
-          var multiSel = this.columns.length > this.cells.length ? 1 : 0;
-
-          for (var i = 0, l = this.cells.length, colspan; i < l; i += colspan) {
+        var size = this.cells.length;
+        if (this.columns.length == size) {
+          for (var i = 0, colspan; i < size; i += colspan) {
             var cell = this.cells[i];
             colspan = cell.colspan || 1;
-            var width = this.columns[i + multiSel].width;
-            var flex = this.columns[i + multiSel].flex;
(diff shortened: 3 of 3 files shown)


tomivirkki:
Don't add asynchronicity to the component. As you see, many tests would require async with this change (same would be required from anyone using the Grid), which is not something we want.

Rather find the reason the API is called so often and fix the original issue.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-:-KSjpIY4TUoYg_kvG6_S:b-8j5cju)*

tomivirkki:
Furthermore, this isn't probably the best time to do optimization since the internals are still subject to heavy change; "premature optimization is the root of all evil".

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-:-KSjpvSiDtYivamUOvk8:b-vh4qzk)*

Saulis:
In this case there are two places where `_updateCells` is called. The other one is the `colspan` listener for each `cell`, which we might either add a default value or remove altogether during #441 and the other place is here: https://github.com/vaadin/vaadin-grid/blob/2.0-dev/vaadin-grid-table-row.html#L155. As you can see, there's already a mention about optimization, but that should be taken care of separately.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-:-KSjpijGYAhKvv_KSa-s:bbvcdp7)*

manolo:
@tomivirkki for instance when the grid is being configured, adding all columns, each single configuration step produces a call to this

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-:-KSju4tvz_1A7YdTdlNP:bat5pr7)*

tomivirkki:
@manolo For sure. The bottom line is that the tests should pass unchanged. We don't want to require async() from the tests and especially from the users.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-:-KSjv-rs74HHnbdpiKyo:b-2gyshf)*

manolo:
Removed the debounce. 
The PR is more a refactoring than an optimisation, it makes easier to understand what the code does (at least to me).
Also a fix when number of columns and cells are different.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-:-KSkCJViwuHOdC1NxTmG:b-vfe89q)*

tomivirkki:
*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkIu_l5mxMrVTFGLIM:b12hmir) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> ```HTML
>       _updateCells: function() {
>         // columns > cells means that there is an extra col for multiselection
>         var multiSel = this.columns.length > this.cells.length ? 1 : 0;
> ```

Looks a lot better now. The hierarchy support will add another extra column so could this be reformatted to
`var extraColumns = this.columns.length - this.cells.length;`
to make this future proof. The variable would also be agnostic to a specific feature.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkMm1q0qQLC_YIXRKi:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> Looks a lot better now. The hierarchy support will add another extra column so could this be reformatted to
> `var extraColumns = this.columns.length - this.cells.length;`
> to make this future proof. The variable would also be agnostic to a specific feature.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkOvW_jg4J26iu1wfb:bbng7iu) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> Done.
</details>
Actually it seems that the variable isn't needed at all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkR6kGbfx48R4UbipI:bin6qcj) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> Actually it seems that the variable isn't needed at all.
</details>
what do you mean ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkRCCssuPICDzsSy18:bjs459x) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> what do you mean ?
</details>
There's a moment between adding extra columns and changing the row.cells property when the extraColumns isn't 0 but that changes soon after related observers are invoked. This is another issue and not critical to fix now.

So practically the extraColumns can always be 0 -> it may as well be removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkSeOPTm7WyEmnlBow:b-5fak7y) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> There's a moment between adding extra columns and changing the row.cells property when the extraColumns isn't 0 but that changes soon after related observers are invoked. This is another issue and not critical to fix now.
> 
> So practically the extraColumns can always be 0 -> it may as well be removed.
> 
</details>
The problem is that sometimes we are setting erroneous properties to the wrong cell.
When debugging is quite confusing, and it works by chance, because always happens that the correct configuration comes the last.
I would do either: return when both arrays have not the same size, or leave the code with the offset check. But would not set erroneous css properties (that could have subtle rendering performance hits)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkUAKN1zpbAkB-b83E:bk37ip5) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> The problem is that sometimes we are setting erroneous properties to the wrong cell.
> When debugging is quite confusing, and it works by chance, because always happens that the correct configuration comes the last.
> I would do either: return when both arrays have not the same size, or leave the code with the offset check. But would not set erroneous css properties (that could have subtle rendering performance hits)
> 
</details>
Well, as I said, "this is another issue and not critical to fix now". But sure you can return if the lengths don't match.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-table-row.html, line 103 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkIu_l5mxMrVTFGLIL:-KSkk3IvL9QAhmRRCTLs:b-9u299q) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9262dc0c2b272a5134e78d305a7edbc7efd94052/vaadin-grid-table-row.html#L103)):*
> Well, as I said, "this is another issue and not critical to fix now". But sure you can return if the lengths don't match.
</details>
Done, not doing anything when length does not match

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

Saulis:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSkpnTgnUvFIXsR6HHd:b-tsh0f4) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> ```HTML
> 
>             // Visit colspanned cells to hide them, and increase parent width and flex.
>             for (var j = 1; j < colspan && i + j < l; j++) {
> ```

Could this be `var j =  i + 1; j < i + colspan && j < l; j++` ? All the consequent `i + j` could then be changed to just `j` 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSkqXA83C63cy50Taxu:bqa5wku) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> Could this be `var j =  i + 1; j < i + colspan && j < l; j++` ? All the consequent `i + j` could then be changed to just `j` 
</details>
Could also remove the l variable. colspan property will not be changeable by the developer. Column-groups are in control of the colspan value internally so there's no need for the check for invalid (overflowing) value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoZVvcLHOjasAzwmf7:b8r0n0v) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> Could also remove the l variable. colspan property will not be changeable by the developer. Column-groups are in control of the colspan value internally so there's no need for the check for invalid (overflowing) value.
</details>
I know that it does not mutate, but the l variable is for simplicity, it's used in a couple of places, and leaving the l by the long  this.cells.length in the second condition does not simplifies the readability,  I think

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSocR9MLCjFkLLv1fky:b-1ngw3x) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> I know that it does not mutate, but the l variable is for simplicity, it's used in a couple of places, and leaving the l by the long  this.cells.length in the second condition does not simplifies the readability,  I think
</details>
Sorry, I misunderstood. Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoc4KpsWU-m81hueqc:bs932c6) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> I know that it does not mutate, but the l variable is for simplicity, it's used in a couple of places, and leaving the l by the long  this.cells.length in the second condition does not simplifies the readability,  I think
</details>
I really don't think that the current approach is simpler than removing the variable. Take a look https://www.dropbox.com/s/g6vx6grmgyxhdf2/Screenshot%202016-09-29%2009.24.56.png?dl=0

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSocXrtlslJK3dsh0aq:b-4jtorm) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> I really don't think that the current approach is simpler than removing the variable. Take a look https://www.dropbox.com/s/g6vx6grmgyxhdf2/Screenshot%202016-09-29%2009.24.56.png?dl=0
</details>
Ok, will take a look

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSod1cDLbZYiUbHURIZ:bfcjfp6) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> Ok, will take a look
</details>
We cannot remove the check, if the user sets a big colspan, say 5 in a 4 columns grid, it breaks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSodKnLaFK75CkMyRyR:b-7nwgqx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> We cannot remove the check, if the user sets a big colspan, say 5 in a 4 columns grid, it breaks
</details>
User isn't in control of the colspan. It's column-group's job.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSodYpgNC_nxkkDAfM4:b2b6rha) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> User isn't in control of the colspan. It's column-group's job.
</details>
I don't follow, this breaks the grid:

```
         <vaadin-grid-column >
            <template>
              <div>baz-[[index]]</div>
            </template>
            <template is="header">
              baz-h1
            </template>
            <template is="header" colspan="5">
              baz-h2
            </template>
          </vaadin-grid-column>

```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSodfUWhoxtLDvut1r3:bnezcze) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> I don't follow, this breaks the grid:
> 
> ```
>          <vaadin-grid-column >
>             <template>
>               <div>baz-[[index]]</div>
>             </template>
>             <template is="header">
>               baz-h1
>             </template>
>             <template is="header" colspan="5">
>               baz-h2
>             </template>
>           </vaadin-grid-column>
> 
> ```
</details>
`colspan` is temporary dev-time property. It will be removed when column-groups are introduced.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSodqj_8rgRfHa5EFbk:bnaq0bz) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> `colspan` is temporary dev-time property. It will be removed when column-groups are introduced.
</details>
Meaning that it can never have an erroneous value since user cannot change it himself.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSodr0m4cCvJyrnpzzh:b-8se1gk) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> `colspan` is temporary dev-time property. It will be removed when column-groups are introduced.
</details>
But it's out of the scope of this PR when it will be removed. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoe0eQNhXh8ChPpFoL:b-2gcnz3) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> But it's out of the scope of this PR when it will be removed. 
</details>
So is adding the check. It wasn't in the previous implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoe798oyU3sLzcb6cc:bmee4tf) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> So is adding the check. It wasn't in the previous implementation.
</details>
Previous implementation was failing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoeHobxi5DOhlsvzAp:b-lbut5x) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> Previous implementation was failing
</details>
What do you mean?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoeMwTSH3SqdfSyvjN:bwthpcp) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> What do you mean?
</details>
In current 2.0-dev if user sets a colspan of 5 in the demo.html, the code breaks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSofGETZiurquG9PyYI:bu2xc4g) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> In current 2.0-dev if user sets a colspan of 5 in the demo.html, the code breaks.
</details>
Ok, if you're going to add this check then please add tests for it also.

Also check that the colspan cannot overflow the frozen column border (1 frozen column with colspan of 2 for example) because that's illegal also and "was failing" in the previous implementation. It also needs tests.

Then the colspan can also be set to -1 by the user which fails too...

To be honest, I don't see a point in adding these checks for a temporary implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSogo_2miqbm_y8dBrI:b-hjs3gy) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> Ok, if you're going to add this check then please add tests for it also.
> 
> Also check that the colspan cannot overflow the frozen column border (1 frozen column with colspan of 2 for example) because that's illegal also and "was failing" in the previous implementation. It also needs tests.
> 
> Then the colspan can also be set to -1 by the user which fails too...
> 
> To be honest, I don't see a point in adding these checks for a temporary implementation.
</details>
Well, I need this for the grouping prototyping I'm doing.
Introducing small changes and refactoring is not as bad ;-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoi8A5sl0a2VcSgP4N:bi9z0py) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> Well, I need this for the grouping prototyping I'm doing.
> Introducing small changes and refactoring is not as bad ;-)
</details>
Well then it should be in your prototyping branch, not in 2.0-dev

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSoiaqAdc3SBlDqovKF:b-xfptd0) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> Well then it should be in your prototyping branch, not in 2.0-dev
</details>
This is related to what @saulis  asked when I started with this story : `The first goal to get done is to modify https://github.com/vaadin/vaadin-grid/blob/2.0-dev/vaadin-grid-table-row.html#L101 to just get the group header cell width calculated more simpler`

I did the PR in order that he could take a look to my findings and refactoring.

I don't think this is the definitive implementation, as everything I do. I still believe in agile development, that means refactor whenever something could be improved.

To be honest, If your thoughts were that this wasn't never to be merged, it could be said at the beginning.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

Saulis:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSokgB-YpFZwaZGcbpi:bv55u6g) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> This is related to what @saulis  asked when I started with this story : `The first goal to get done is to modify https://github.com/vaadin/vaadin-grid/blob/2.0-dev/vaadin-grid-table-row.html#L101 to just get the group header cell width calculated more simpler`
> 
> I did the PR in order that he could take a look to my findings and refactoring.
> 
> I don't think this is the definitive implementation, as everything I do. I still believe in agile development, that means refactor whenever something could be improved.
> 
> To be honest, If your thoughts were that this wasn't never to be merged, it could be said at the beginning.
> 
> 
</details>
@tomivirkki isn't against this refactoring, he's just stating that if you're going to change the behavior, it's not refactoring anymore and you should add tests for the changed behavior. In this case, there hasn't been a check for overflowing colspan by the user, and at this point we probably shouldn't add one since the grouping feature will drop that from the API. So I agree in that sense that it might be better to leave the whole check out at this point.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
*[vaadin-grid-table-row.html, line 110 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471#-KSkpnTgnUvFIXsR6HHc:-KSolLRzohXJHj_PLpnd:b4p9b4k) ([raw file](https://github.com/vaadin/vaadin-grid/blob/00aed5e8a2657f9a2d6288eb6b971c852280d3bb/vaadin-grid-table-row.html#L110)):*
> @tomivirkki isn't against this refactoring, he's just stating that if you're going to change the behavior, it's not refactoring anymore and you should add tests for the changed behavior. In this case, there hasn't been a check for overflowing colspan by the user, and at this point we probably shouldn't add one since the grouping feature will drop that from the API. So I agree in that sense that it might be better to leave the whole check out at this point.
</details>
The branch look really good to me. It simplifies the code a lot and I want to merge it but as @saulis stated, the check is unnecessary.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

tomivirkki:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/471)*

PR #476, opened 2016-10-03
Clear selection on refresh items

### LINKED ISSUE #344: Selecting single result in a filtered grid causes entire grid to get selected
If you filter down the contents of a grid to a single item and select that item with multi select mode activated, it will incorrectly select the entire grid.

Steps to reproduce: 
1. Go to https://vaadin.com/elements/-/element/vaadin-grid > Filtering demo
2. Type in 'aa' to narrow down contents to one row
3. Select row - observe that the select all checkbox gets selected
4. Clear filter - observe that entire grid is now selected

Saulis:
Fixes #344 

When `refreshItems()` is called, data is refetched from the data source. As we have no guarantee that the items returned will be the same and/or in same order as before, there's a high possibility that `selected` indexes are incorrect. This leads to situations like described in #344.

In some cases clearing the selection on refresh might be undesirable, if we won't do it by default, the alternative solution would be to add documentation and instruct users to clear the selection manually before calling `refreshItems()`.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/476)

### DIFF
--- a/test/grid-binding-data.html
+++ b/test/grid-binding-data.html
@@ -282,6 +282,14 @@
           expect(qLocal('.vaadin-grid-body .vaadin-grid-cell')).to.exist;
         });
       });
+
+      it('should clear selection', function() {
+        grid.selection.select(0);
+
+        grid.refreshItems();
+
+        expect(grid.selection.selected()).to.be.empty;
+      });
     });
 
     describe('size', function() {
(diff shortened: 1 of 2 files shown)


platosha:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/476#-:-KT9mfI9omBpXYQhoMT7:bnfp4nl)*

PR #481, opened 2016-10-07
Make sure _ignoreScroll gets cleared

tomivirkki:
If scroller._ignoreScroll is true while resizing the grid, the headers/footers might get misplaced. This PR makes sure  _ignoreScroll flag gets cleared.
screenshot 2016-10-07 12 00 00

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/481)

### DIFF
--- a/test/helpers.html
+++ b/test/helpers.html
@@ -44,7 +44,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   function scrollToEnd(grid, callback) {
     var table = grid.$.scroller.$.table;
-    table.addEventListener('scroll', function () {
+    var listener = function () {
       grid.async(function () {
         table.scrollTop = table.scrollHeight;
       }, 1);
@@ -52,9 +52,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       grid.debounce('scroll-done', function () {
         // Ensure rows are in order
         grid.flushDebouncer('reorderrows');
+        table.removeEventListener('scroll', listener);
         callback();
       }, 100);
-    });
+    };
+    table.addEventListener('scroll', listener);
     table.scrollTop = table.scrollHeight;
   }
 
--- a/test/outer-scroller.html
+++ b/test/outer-scroller.html
@@ -51,11 +51,11 @@
         expect(spy.called).to.be.true;
       });
 
-      it('should invoke frozen element translate handler once', function(done) {
+      it('should invoke frozen element translate handler', function(done) {
         var spy = sinon.spy(scroller, '_translateStationaryElements');
         outerScroller.scrollLeft = 1;
         setTimeout(function () {
-          expect(spy.callCount).to.equal(1);
+          expect(spy.called).to.be.true;
           done();
         }, 200);
       });
--- a/test/row-height.html
+++ b/test/row-height.html
@@ -209,6 +209,45 @@
         };
       });
 
+      it('should not misalign on resize (wheel)', function(done) {
+        init('default-content');
+        scrollToEnd(grid, function() {
+          // Simulate a wheel event that doesn't cause a scroll event
+          var e = new CustomEvent('wheel', {bubbles: true});
+          e.deltaY = 0;
+          e.deltaX = -100;
+          scroller.$.table.dispatchEvent(e);
+
+          // Simulate a (window) resize
+          grid.style.width = '200px';
+          scroller.fire('iron-resize');
+
+          scroller.async(function() {
+            expect(header.getBoundingClientRect().top).to.be.closeTo(scroller.getBoundingClientRect().top, 1);
+            done();
+          }, 100);
+        });
+      });
+
+      it('should not misalign on resize (outer-scroller)', function(done) {
+        init('default-content');
+        scrollToEnd(grid, function() {
+          scroller.$.outerscroller.addEventListener('scroll', function() {
+            // Simulate a (window) resize
+            grid.style.width = '200px';
+            scroller.fire('iron-resize');
+
+            scroller.async(function() {
+              expect(header.getBoundingClientRect().top).to.be.closeTo(scroller.getBoundingClientRect().top, 1);
+              done();
+            }, 100);
+          });
+
+          // Scroll with the outer-scroller
+          scroller.$.outerscroller.scrollTop = 100;
+        });
+      });
+
     });
   </script>
 
--- a/vaadin-grid-table-outer-scroller.html
+++ b/vaadin-grid-table-outer-scroller.html
@@ -96,7 +96,6 @@
           this.scrollTarget.scrollLeft = this.scrollLeft;
 
           this.domHost._scrollHandler();
-          this.domHost._ignoreScroll = true;
         }
         this._syncingOuterScroller = false;
 
--- a/vaadin-grid-table-scroll-behavior.html
+++ b/vaadin-grid-table-scroll-behavior.html
@@ -78,7 +78,6 @@
       this.$.table.scrollTop += e.deltaY;
       this.$.table.scrollLeft += e.deltaX;
       this._scrollHandler();
-      this._ignoreScroll = true;
     },
 
     _adjustVirtualIndexOffset: function(delta) {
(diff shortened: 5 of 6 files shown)


Saulis:
The `_ignoreScroll` is only for optimizing, right? What if we would just remove the optimization to clean up the code a bit? Didn't test on IE or Edge, but on browsers I can't tell any difference performance wise.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KThloWG6vxrjvpWMfJy:b-ns2z2q)*

tomivirkki:
It is for optimizing only. I'd assume the performance impact to be quite big though (might not be visible on desktop) as all the logic that's run on scroll would be run twice without it. If there's no notable difference on mobile devices then +1 for removing it.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KThp9LPOATRm24uMwV9:b-k04bqm)*

Saulis:
Mobile devices don't set the `ignoreScroll` to `true` as they don't fire `wheel` events right?

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTi-AyewkcPqmHd9L8M:b-4ar4wi)*

tomivirkki:
They still use the outer-scroller which does the same thing. But what I mostly meant was that we should test with something less performant than your super-laptop :)

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTi-XarTHpntdrlJ7rh:bh7mu54)*

Saulis:
Oh right, I'd totally missed that outer scroller also uses the property.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTiDIG69CV72VXMhjoG:b-hn5kp3)*

manolo:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/row-height.html, line 230 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-KTmBSWbGDswVUNY00L7:-KTmBSWbGDswVUNY00L8:b-xhuped) ([raw file](https://github.com/vaadin/vaadin-grid/blob/d6934847d20dfbb33288c9168e6466dae95a23f0/test/row-height.html#L230)):*
> ```HTML
>               done();
>             }, 100);
>           }, 20); // Debounce perior for 'clear-ignorescroll' is 10
> ```

perior -> period  ?

Shouldn't  call flushDebounced instead of wait for it ? 

---

*[test/row-height.html, line 252 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-KTmBNA9zgL5snJo9Rx0:-KTmBNA9zgL5snJo9Rx1:b3rfc8w) ([raw file](https://github.com/vaadin/vaadin-grid/blob/d6934847d20dfbb33288c9168e6466dae95a23f0/test/row-height.html#L252)):*
> ```HTML
>         });
>       });
>       
> ```

nit : some spaces left

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481)*

tomivirkki:
*[test/row-height.html, line 230 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-KTmBSWbGDswVUNY00L7:-KTmOqvLz0-OUSV4vBK3:bjv50mc) ([raw file](https://github.com/vaadin/vaadin-grid/blob/d6934847d20dfbb33288c9168e6466dae95a23f0/test/row-height.html#L230)):*
> perior -> period  ?
> 
> Shouldn't  call flushDebounced instead of wait for it ? 
</details>
Fixed the typo.

The timeout is needed as the async 'scroll' event might still be on the way even if we flushed the debouncer.

Also I think this test shouldn't flush the debouncer that the same PR introduces; from TDD point of view the debouncer doesn't exist (before the fix).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481)*

tomivirkki:
*[test/row-height.html, line 252 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-KTmBNA9zgL5snJo9Rx0:-KTmPRmAst-n2gccfTQR:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/d6934847d20dfbb33288c9168e6466dae95a23f0/test/row-height.html#L252)):*
> nit : some spaces left
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481)*

Saulis:
OK how about this: can we drop listening for `scroll` events altogether on the table, if we move the `wheel` listener to the grid element? Then all scrolling for the table should happen either through the `wheel` event handler on the grid or the outer scroller? That would also allow dropping the scroll syncing from table -> outer-scroller, right?

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTmXbMnGoUH95O-NKGX:bxd1dul)*

Saulis:
What I kinda mean is that `wheel` events would scroll the outer-scroller instead of the table.

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTmYmxMwRuyou7mLJGI:b-djkws2)*

tomivirkki:
The outer-scroller isn't in use in all cases. In some environments it's hidden on purpose.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTmZKCXav5ZAVI26_Ib:bawa9ed)*

Saulis:
Ah, OK.  I did do some perf testing on different devices (old and new ipad we have, nexus phone, surface) using the `<fps-meter>`, there is a 1-2 FPS difference depending on the device, but nothing noticeable. We probably should figure out an automated test to benchmark the performance somehow.

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTmiPd-OAWAD6qIn6Ys:b-7p0nfr)*

tomivirkki:
Ok, sounds good. Changed the PR so that it removes _ignoreScroll optimization.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/row-height.html, line 230 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-KTmBSWbGDswVUNY00L7:-KTmumUFkS4bNgZgx2NU:b-1urd85) ([raw file](https://github.com/vaadin/vaadin-grid/blob/d6934847d20dfbb33288c9168e6466dae95a23f0/test/row-height.html#L230)):*
> Fixed the typo.
> 
> The timeout is needed as the async 'scroll' event might still be on the way even if we flushed the debouncer.
> 
> Also I think this test shouldn't flush the debouncer that the same PR introduces; from TDD point of view the debouncer doesn't exist (before the fix).
</details>
Fixed, the flag was removed so no need to wait > the debounce period

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTmui6LqGryzamctzgG:bs3ysfv)*

Saulis:
:lgtm:

---

Reviewed 1 of 3 files at r1, 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/481#-:-KTnpt-eOI7IssL1Qq2X:bnfp4nl)*

PR #488, opened 2016-10-13
No individual shadow roots for each separate cell, better content selectors

tomivirkki:
These fixes mainly target grid with native shadow DOM on:
- Each cell doesn't need to have individual shadow roots (just the content selector)
- Class names aren't the right choice for content selectors in this case, using attributes
- Light dom items shouldn't accumulate on dom changes (for example when number of columns change)
- vaadin-grid-table-cell implementation is now more in line with other vaadin-grid-table- components

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/488)

### DIFF
--- a/vaadin-grid-table-cell.html
+++ b/vaadin-grid-table-cell.html
@@ -1,48 +1,6 @@
-<dom-module id="vaadin-grid-table-cell-styles">
-  <template>
-    <style>
-       ::content>div {
-        padding: 8px;
-        width: 100%;
-        box-sizing: border-box;
-        display: flex;
-        justify-content: center;
-        flex-direction: column;
-        white-space: nowrap;
-      }
-
-       :host {
-        flex-shrink: 0;
-        flex-grow: 1;
-        flex-basis: 100px;
-        box-sizing: border-box;
-        display: flex;
-        overflow: hidden;
-      }
-    </style>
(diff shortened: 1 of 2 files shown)


Saulis:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[vaadin-grid-table-cell.html, line 126 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/488#-KU1jwsmGzKNXzqNW-bu:-KU1jwsmGzKNXzqNW-bv:bbm3uj2) ([raw file](https://github.com/vaadin/vaadin-grid/blob/0d534ef73a195271324e9033d201297050179d60/vaadin-grid-table-cell.html#L126)):*
> ```HTML
>           // not helping here for some reason.
>           this.async(function() {
>             if (this.target.domHost) {
> ```

Comment: looks like this logic doesn't work perfectly when grid is nested inside a `dom-bind` template, `domHost` is then null. However, `vaadin-grid` seems to be assigned with a `x-scope` class. I'll file a separate issue about that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/488#-:-KU1kMbyaCpcT8ZlMpRC:bnfp4nl)*

PR #493, opened 2016-10-14
Add Column Groups

Saulis:
Connects to #441

This PR doesn't have the `frozen` functionality yet, adding that separately.

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/493)


Saulis:
ping @magi42

tomivirkki:
*[docs/vaadin-grid-assigning-data.adoc, line 30 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KU8EQc-DoJ3DLPlVJDu:-KU8EQc-DoJ3DLPlVJDv:b-h3jorx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/4fa5210e5ecc7fe8a2f185d951b6b9cc691bfbbb/docs/vaadin-grid-assigning-data.adoc#L30)):*
> ```AsciiDoc
> <vaadin-grid>
>   <vaadin-grid-column>
>     <template is="header">First</template>
> ```

The "is" attribute is reserved for customized built-in elements as [documented in the custom elements spec](https://www.w3.org/TR/custom-elements/#customized-built-in-element) (examples: "dom-repeat" or "iron-input").

Since "header" isn't such an extension (or even a valid custom element name) we shouldn't use "is" here for marking the header template either.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
Review status: 0 of 26 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[vaadin-grid-column-group.html, line 98 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KUw3gzrzoSgSuVyYYS7:-KUw3gzrzoSgSuVyYYS8:b-4811so) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b591c6b0c1f6dd58700ca4a32127fb8a0c03227a/vaadin-grid-column-group.html#L98)):*
> ```HTML
>           path: 'headerTemplate',
>           value: template
>         }, {bubbles: false});
> ```

Why bubbles: false, tests pass without it and demo looks ok

---

*[vaadin-grid-column-group.html, line 3 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-T2708qMreCsHh5CL:-KV-T2708qMreCsHh5CM:b8uxwyt) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-column-group.html#L3)):*
> ```HTML
> <link rel="import" href="../polymer/polymer.html">
> 
> <dom-module id="vaadin-grid-column-group">
> ```

`vaadin-grid-column-group` looks like an extension to `vaadin-grid-column`. It's handled as a 'column' further in the code and they share a lot of codebase... Almost calls for a `VaadinGridColumnBehavior`.

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV-U7RDk9sVlTCrBDj1:bauj41q) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> ```HTML
> 
>       td[empty], th[empty] {
>         border-bottom: none;
> ```

A couple of things regarding the groups appearance:
 - Looks like some cells are missing left/right paddings (only have underlining border).
 - If a column is missing a header template ('empty' cell at the last header row), there's no border between header and body.

---

*[vaadin-grid.html, line 100 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-SnnM4r1W39kGp505:-KV-SnnM4r1W39kGp506:b-ts8fog) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid.html#L100)):*
> ```HTML
>       }
> 
>       this.columns = columns;
> ```

`columns` should be a private property right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
> 540e834

Latest commit  breaks test.  I don't fully understand what is this commit for. It does not have any unit test either.

---

Review status: 0 of 26 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-:-KV1Ko8SRU9dfVmLlAU_:b-r6arlr)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-grid.html, line 100 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-SnnM4r1W39kGp505:-KV42zUZBMDVksGnSSRb:b-86e5cv) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid.html#L100)):*
> `columns` should be a private property right?
</details>
Not sure, this is the only way to modify the grid dynamically right now, for instance to change frozen, width etc

---

*[docs/vaadin-grid-assigning-data.adoc, line 30 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KU8EQc-DoJ3DLPlVJDu:-KV43DzEhVLk8Df2JEnI:b-r1kz7r) ([raw file](https://github.com/vaadin/vaadin-grid/blob/4fa5210e5ecc7fe8a2f185d951b6b9cc691bfbbb/docs/vaadin-grid-assigning-data.adoc#L30)):*
> The "is" attribute is reserved for customized built-in elements as [documented in the custom elements spec](https://www.w3.org/TR/custom-elements/#customized-built-in-element) (examples: "dom-repeat" or "iron-input").
> 
> Since "header" isn't such an extension (or even a valid custom element name) we shouldn't use "is" here for marking the header template either.
</details>
I think we could use just an attribute `<template header>` better than a class wdyt ? 

---

*[vaadin-grid-column-group.html, line 98 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KUw3gzrzoSgSuVyYYS7:-KV43yI4MMu53KNEc2R9:b653jfh) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b591c6b0c1f6dd58700ca4a32127fb8a0c03227a/vaadin-grid-column-group.html#L98)):*
> Why bubbles: false, tests pass without it and demo looks ok
</details>
Just to avoid it being processed up, since it's not used. Not a big performance improvement.

---

*[vaadin-grid-column-group.html, line 3 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-T2708qMreCsHh5CL:-KV43VsGaBTZru21Jj3P:bb6un1j) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-column-group.html#L3)):*
> `vaadin-grid-column-group` looks like an extension to `vaadin-grid-column`. It's handled as a 'column' further in the code and they share a lot of codebase... Almost calls for a `VaadinGridColumnBehavior`.
</details>
Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
Review status: 0 of 26 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-grid.html, line 100 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-SnnM4r1W39kGp505:-KV45_chnQwEnhptgfSw:bdk2ahz) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid.html#L100)):*
> Not sure, this is the only way to modify the grid dynamically right now, for instance to change frozen, width etc
</details>
Frozen/width/order/hierarchy are all configured trough the DOM elements.

---

*[docs/vaadin-grid-assigning-data.adoc, line 30 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KU8EQc-DoJ3DLPlVJDu:-KV4623crEorUq82EszN:b-f87g7o) ([raw file](https://github.com/vaadin/vaadin-grid/blob/4fa5210e5ecc7fe8a2f185d951b6b9cc691bfbbb/docs/vaadin-grid-assigning-data.adoc#L30)):*
> I think we could use just an attribute `<template header>` better than a class wdyt ? 
</details>
Well, "header" is a non-standard attribute for a `<template>` so i'd avoid using it (especially in public API)

---

*[vaadin-grid-column-group.html, line 98 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KUw3gzrzoSgSuVyYYS7:-KV46cDNtjS92OfDpFa5:b-dkgst7) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b591c6b0c1f6dd58700ca4a32127fb8a0c03227a/vaadin-grid-column-group.html#L98)):*
> Just to avoid it being processed up, since it's not used. Not a big performance improvement.
</details>
Rest of the "property-changed" -events are still bubbling outside the grid? Maybe align these (catch 'em all at grid level and stop propagation for example...or just let all of them bubble)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-column.html, line 100 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV47WtU9hNE8XpEAzsX:-KV47WtU9hNE8XpEAzsY:b-f9y0fg) ([raw file](https://github.com/vaadin/vaadin-grid/blob/18a01016ff3e510a0c0f9999480e7d79c0e7bac5/vaadin-grid-column.html#L100)):*
> ```HTML
>       behaviors: [vaadinGridColumnBehavior],
> 
>       properties: {
> ```

All of these properties are already defined by the behavior

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid-column-group.html, line 98 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KUw3gzrzoSgSuVyYYS7:-KV47mQeJHo1HA2d1Dfy:bpd5qir) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b591c6b0c1f6dd58700ca4a32127fb8a0c03227a/vaadin-grid-column-group.html#L98)):*
> Rest of the "property-changed" -events are still bubbling outside the grid? Maybe align these (catch 'em all at grid level and stop propagation for example...or just let all of them bubble)?
</details>
Already stoping them at this point but in a separate PR. I don't see any problem of not bubbling these, but if you do prefer have all events aligned there is no problem in my part

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4K8PalTxjzXncNlfV:b49cv9y) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> A couple of things regarding the groups appearance:
>  - Looks like some cells are missing left/right paddings (only have underlining border).
>  - If a column is missing a header template ('empty' cell at the last header row), there's no border between header and body.
</details>
Can you send a screenshot, cannot figure out what you mean

---

*[vaadin-grid.html, line 100 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-SnnM4r1W39kGp505:-KV4KIJ2wmAx2UREEfvJ:befi7l1) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid.html#L100)):*
> Frozen/width/order/hierarchy are all configured trough the DOM elements.
</details>
I wanted to avoid the user to have to transverse the dom to find a specific group/column. The arrays of arrays in columns already have done some work to create a matrix of the headers in their appropriate position.

---

*[docs/vaadin-grid-assigning-data.adoc, line 30 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KU8EQc-DoJ3DLPlVJDu:-KV4Ko83pnknNzQTeKew:b-p55mua) ([raw file](https://github.com/vaadin/vaadin-grid/blob/4fa5210e5ecc7fe8a2f185d951b6b9cc691bfbbb/docs/vaadin-grid-assigning-data.adoc#L30)):*
> Well, "header" is a non-standard attribute for a `<template>` so i'd avoid using it (especially in public API)
</details>
Seems quite common in polymer to use additional attributes, even in template for instance `template is="dom-repeat" items...` 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

Saulis:
Review status: 0 of 26 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-grid-column.html, line 136 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4IUVXaGFOZuroTSYs:-KV4IUVXaGFOZuroTSYt:bhe3n3h) ([raw file](https://github.com/vaadin/vaadin-grid/blob/18a01016ff3e510a0c0f9999480e7d79c0e7bac5/vaadin-grid-column.html#L136)):*
> ```HTML
>           }
>         },
>         _flex: Number,
> ```

`_flex` and `_width` need to remove now when the behavior is providing `flex` and `width`

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4JQ0Zn8ul8hdlFrOW:b-hhe9mx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> A couple of things regarding the groups appearance:
>  - Looks like some cells are missing left/right paddings (only have underlining border).
>  - If a column is missing a header template ('empty' cell at the last header row), there's no border between header and body.
</details>
Hmm can you show me an example about the paddings?

About the missing borders – yeah, not sure what to do about it – if we remove the `:not([empty])` from the selectors from above , for example `tfoot` will always have a 1px border even if there aren't any footer templates defined.

---

*[vaadin-grid.html, line 100 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-SnnM4r1W39kGp505:-KV4K7pSQ8k_70dmb_4q:b-uxq5op) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid.html#L100)):*
> Frozen/width/order/hierarchy are all configured trough the DOM elements.
</details>
yeah we should make it private at this point. Would be OK to do after this has been merged?

---

*[docs/vaadin-grid-assigning-data.adoc, line 30 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KU8EQc-DoJ3DLPlVJDu:-KV4KTwpYKwT_F83m93M:b-np2fyo) ([raw file](https://github.com/vaadin/vaadin-grid/blob/4fa5210e5ecc7fe8a2f185d951b6b9cc691bfbbb/docs/vaadin-grid-assigning-data.adoc#L30)):*
> Well, "header" is a non-standard attribute for a `<template>` so i'd avoid using it (especially in public API)
</details>
I'm cool with using a class, but as it's a bit out of scope for this PR, would it be OK to do separately?

---

*[vaadin-grid-column-group.html, line 98 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KUw3gzrzoSgSuVyYYS7:-KV4KmF6chA0suEYupgW:b-i5004n) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b591c6b0c1f6dd58700ca4a32127fb8a0c03227a/vaadin-grid-column-group.html#L98)):*
> Already stoping them at this point but in a separate PR. I don't see any problem of not bubbling these, but if you do prefer have all events aligned there is no problem in my part
</details>
Yeah, the frozen columns PR will have code in table level which will stop the events from bubbling. `bubbles: false` on the header and footer templates actually have  a reason, I'll make  a test for that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4LYHiKiLL0w2Cz3oO:bve71m9) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> Hmm can you show me an example about the paddings?
> 
> About the missing borders – yeah, not sure what to do about it – if we remove the `:not([empty])` from the selectors from above , for example `tfoot` will always have a 1px border even if there aren't any footer templates defined.
</details>
@manolo here's a screenshot (grp.html without the demo styles) https://www.dropbox.com/s/7m0kki1o2n7w3rt/Screenshot%202016-10-27%2011.55.27.png?dl=0

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid.html, line 100 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-SnnM4r1W39kGp505:-KV4M-qSK5BIwiBuNTAk:b-3i1t1) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid.html#L100)):*
> yeah we should make it private at this point. Would be OK to do after this has been merged?
</details>
Sure, works for me

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4M--V33LAJQGtCCWN:b-uphggy) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> @manolo here's a screenshot (grp.html without the demo styles) https://www.dropbox.com/s/7m0kki1o2n7w3rt/Screenshot%202016-10-27%2011.55.27.png?dl=0
</details>
I would remove all borders, but the header bottom and footer top (like in 1.x) , and apply a couple of mixins and leave the user do the work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[docs/vaadin-grid-assigning-data.adoc, line 30 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KU8EQc-DoJ3DLPlVJDu:-KV4MSgHUZtvqA9095ve:bk3crt) ([raw file](https://github.com/vaadin/vaadin-grid/blob/4fa5210e5ecc7fe8a2f185d951b6b9cc691bfbbb/docs/vaadin-grid-assigning-data.adoc#L30)):*
> I'm cool with using a class, but as it's a bit out of scope for this PR, would it be OK to do separately?
</details>
@manolo "items" is defined by "dom-repeat" so it's valid.
@saulis Ok to do it in a separate PR

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid-column.html, line 100 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV47WtU9hNE8XpEAzsX:-KV4NF-R85_OT4ZGPACf:bb6un1j) ([raw file](https://github.com/vaadin/vaadin-grid/blob/18a01016ff3e510a0c0f9999480e7d79c0e7bac5/vaadin-grid-column.html#L100)):*
> All of these properties are already defined by the behavior
</details>
Done

---

*[vaadin-grid-column.html, line 136 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4IUVXaGFOZuroTSYs:-KV4NIzOV2fPgsO85qvq:be0vjuu) ([raw file](https://github.com/vaadin/vaadin-grid/blob/18a01016ff3e510a0c0f9999480e7d79c0e7bac5/vaadin-grid-column.html#L136)):*
> `_flex` and `_width` need to remove now when the behavior is providing `flex` and `width`
</details>
We want those be private in groups, so they are defined separately in columns and groups and with different names, not shared in the behavior. 
The method is shared, but the variable passed is different because don't share the observers

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

Saulis:
Review status: 0 of 26 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid-column.html, line 136 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4IUVXaGFOZuroTSYs:-KV4OivTEl12vpiJlxDM:bwhtp5h) ([raw file](https://github.com/vaadin/vaadin-grid/blob/18a01016ff3e510a0c0f9999480e7d79c0e7bac5/vaadin-grid-column.html#L136)):*
> We want those be private in groups, so they are defined separately in columns and groups and with different names, not shared in the behavior. 
> The method is shared, but the variable passed is different because don't share the observers
</details>
Yes you're right

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-column.html, line 103 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4OrZGRzU-bfpQQmc4:-KV4OrZGRzU-bfpQQmc5:bvwwtec) ([raw file](https://github.com/vaadin/vaadin-grid/blob/74e08e514539593b8710f99ed79eecd17dbad9e6/vaadin-grid-column.html#L103)):*
> ```HTML
>       },
> 
>       observers: ['_flexChanged(flex)', '_widthChanged(width)', '_templateChanged(template)']
> ```

Should this '_templateChanged(template)' observer be in the behavior also? Now it's not active for a group

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4PKxFz-NfOnzlk-Mu:b-hcy39y) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> I would remove all borders, but the header bottom and footer top (like in 1.x) , and apply a couple of mixins and leave the user do the work.
</details>
@saulis Ah, wrote 'padding' but meant 'border' (the issue in the screenshot)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid-column.html, line 103 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4OrZGRzU-bfpQQmc4:-KV4SKQe30ntKP3U5sHb:b-29i8i3) ([raw file](https://github.com/vaadin/vaadin-grid/blob/74e08e514539593b8710f99ed79eecd17dbad9e6/vaadin-grid-column.html#L103)):*
> Should this '_templateChanged(template)' observer be in the behavior also? Now it's not active for a group
</details>
template is not needed in groups, only columns to notify body cells that it changed. Header and  footer cells use a different approach 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4WJ2t8PoUlDVCvZ7d:b8grbtq) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> @saulis Ah, wrote 'padding' but meant 'border' (the issue in the screenshot)
</details>
Aligned styles with 1.x

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4aVGEDB4toHifnnwM:b-1dl2f8) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> Aligned styles with 1.x
</details>
I don't think the intention was to align the appearance with 1.x. The current dev-2.0 theme is by Jouni. Better off sticking to that until further info.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4bFnT5iXQ2kdr0Xu9:bo26fws) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> I don't think the intention was to align the appearance with 1.x. The current dev-2.0 theme is by Jouni. Better off sticking to that until further info.
</details>
I just wanted not to leave the groups feature with a bad appearance due to borders wrong defined and yet difficult to deal with, until we have more custom props or mixin, so I thought it was ok to remove borders right now. Asked already @jouni: 

`No rush to change it yet, the styles/theme will be revised at some point though. But if you want, you can align the styles with the classic grid.`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4ctc7s8Xg9cY-rCAu:b-h3ihu7) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> I just wanted not to leave the groups feature with a bad appearance due to borders wrong defined and yet difficult to deal with, until we have more custom props or mixin, so I thought it was ok to remove borders right now. Asked already @jouni: 
> 
> `No rush to change it yet, the styles/theme will be revised at some point though. But if you want, you can align the styles with the classic grid.`
</details>
Ok then. Can't target the thead with the box-shadow howerver: https://www.dropbox.com/s/3mw0h47p41741np/Screenshot%202016-10-27%2013.15.53.png?dl=0

Target the last row's cells instead

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4dB-G4mqt1MXYmVoP:b-1eg4j8) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> Ok then. Can't target the thead with the box-shadow howerver: https://www.dropbox.com/s/3mw0h47p41741np/Screenshot%202016-10-27%2013.15.53.png?dl=0
> 
> Target the last row's cells instead
</details>
Use `thead tr:last-child th` as in the previous version

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

Saulis:
Review status: 0 of 26 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-grid-column.html, line 103 at r8](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4OrZGRzU-bfpQQmc4:-KV4dvrD81ezj7vlUf2_:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/74e08e514539593b8710f99ed79eecd17dbad9e6/vaadin-grid-column.html#L103)):*
> template is not needed in groups, only columns to notify body cells that it changed. Header and  footer cells use a different approach 
</details>
Done.

---

*[vaadin-grid-column-group.html, line 98 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KUw3gzrzoSgSuVyYYS7:-KV4dryRJJL3eYREFHcd:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/b591c6b0c1f6dd58700ca4a32127fb8a0c03227a/vaadin-grid-column-group.html#L98)):*
> Yeah, the frozen columns PR will have code in table level which will stop the events from bubbling. `bubbles: false` on the header and footer templates actually have  a reason, I'll make  a test for that.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4ePR_YFWG1H7_ouBo:bjnvb7k) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> Use `thead tr:last-child th` as in the previous version
</details>
Crap, why does it happen?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4gE1IcVQ2n53U8Di2:bgedm7r) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> Crap, why does it happen?
</details>
Because thead is only 100% of the grid's width. There may be cells that extend the scrollWidth but it doesn't affect thead/tfoot

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table.html, line 146 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV-U7RDk9sVlTCrBDj0:-KV4geoUBG5pR3eFrHW1:b-vg1cq) ([raw file](https://github.com/vaadin/vaadin-grid/blob/540e834307717dae5d5934710e963a0ff956d36b/vaadin-grid-table.html#L146)):*
> Because thead is only 100% of the grid's width. There may be cells that extend the scrollWidth but it doesn't affect thead/tfoot
</details>
thx for the explanation. Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 142 at r12](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4zQKV6gfAJNukkLZo:-KV4zQKV6gfAJNukkLZp:b-9l7bmh) ([raw file](https://github.com/vaadin/vaadin-grid/blob/679dfe7e9b2a4413106197769e7d8b0c3f398b61/vaadin-grid-table.html#L142)):*
> ```HTML
> 
>       tbody tr {
>         border-bottom: 1px solid var(--vaadin-grid-border-color);
> ```

Ooh, same issue here, needs to target the cells, not the row
https://www.dropbox.com/s/sio1k2ot3dftgys/Screenshot%202016-10-27%2014.55.42.png?dl=0

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

Saulis:
Review status: 0 of 26 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-grid-table.html, line 142 at r12](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV4zQKV6gfAJNukkLZo:-KV52meNKENXG5Dxylp9:b-886gsx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/679dfe7e9b2a4413106197769e7d8b0c3f398b61/vaadin-grid-table.html#L142)):*
> Ooh, same issue here, needs to target the cells, not the row
> https://www.dropbox.com/s/sio1k2ot3dftgys/Screenshot%202016-10-27%2014.55.42.png?dl=0
</details>
Fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: 0 of 26 files reviewed at latest revision, 4 unresolved discussions.

---

*[vaadin-grid-table.html, line 238 at r13](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV53puUZaebYjoP5v9o:-KV53puUZaebYjoP5v9p:b-d7xeok) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f7cb8d1552385fc04064235dc519f011025ecb87/vaadin-grid-table.html#L238)):*
> ```HTML
>     _columnPropChanged: function(e) {
>       if (/headerTemplate/.test(e.detail.path)) {
>         this.toggleAttribute('has-templates', true, this.$.header);
> ```

Out of curiosity, we do we prefer to use attributes here instead of classes vs the same case with templates? this attribute is not part of any spec or behaviour 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 238 at r13](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV53puUZaebYjoP5v9o:-KV54_mVyf3ymp12sTQ_:b-hz62e2) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f7cb8d1552385fc04064235dc519f011025ecb87/vaadin-grid-table.html#L238)):*
> Out of curiosity, we do we prefer to use attributes here instead of classes vs the same case with templates? this attribute is not part of any spec or behaviour 
</details>
In this case 'this' is a `<vaadin-grid-table>` which is not an existing/default DOM element like ´<template>´ or ´<div>´. Adding custom attributes to existing elements will break validation for the document but that's not the case with custom elements.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
Reviewed 3 of 4 files at r1, 4 of 13 files at r2, 9 of 17 files at r4, 1 of 3 files at r5, 2 of 2 files at r6, 1 of 3 files at r7, 1 of 2 files at r9, 1 of 3 files at r10, 3 of 4 files at r12, 1 of 1 files at r13.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

manolo:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table.html, line 238 at r13](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV53puUZaebYjoP5v9o:-KV58uUTi-jrHvckyZtL:b-j03f0z) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f7cb8d1552385fc04064235dc519f011025ecb87/vaadin-grid-table.html#L238)):*
> In this case 'this' is a `<vaadin-grid-table>` which is not an existing/default DOM element like ´<template>´ or ´<div>´. Adding custom attributes to existing elements will break validation for the document but that's not the case with custom elements.
</details>
I guess validators do not know so much about custom elements, normally they are static parsers.
But anyway, traditionally it was considered that not using class names in selectors would hit the performance, in gwt and jquery they have promoted that for ages. At least old browsers used to perform faster with classnames.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

tomivirkki:
*[vaadin-grid-table.html, line 238 at r13](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493#-KV53puUZaebYjoP5v9o:-KV5BbhlSV77bzz2SMNi:bfbht56) ([raw file](https://github.com/vaadin/vaadin-grid/blob/f7cb8d1552385fc04064235dc519f011025ecb87/vaadin-grid-table.html#L238)):*
> I guess validators do not know so much about custom elements, normally they are static parsers.
> But anyway, traditionally it was considered that not using class names in selectors would hit the performance, in gwt and jquery they have promoted that for ages. At least old browsers used to perform faster with classnames.
</details>
There shouldn't be performance difference in modern browsers and it really isn't the main concern here. This is a question about whether the attribute is valid for the element or not.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/493)*

jouni:
🎉

PR #495, opened 2016-10-18
Fix the count of visible scrollers

tomivirkki:
The grid shows 2 scollbars (on top of each other) (on native shadow) since the `display: block;` defined in `:host` seems to be stronger than `hidden`.

Instead of trying to hide the outer scroller use the same approach as with the table element and set it's overflow hidden to hide the scrollbar.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/495)

### DIFF
--- a/test/helpers.html
+++ b/test/helpers.html
@@ -11,6 +11,23 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <script>
   var defaultCellWidth = 100;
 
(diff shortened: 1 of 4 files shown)


Saulis:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/495)*

PR #504, opened 2016-10-27
Reuse code from columns in group with a behavior

manolo:
Connects to #441

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/504)


manolo:
Bad branch selection

PR #509, opened 2016-10-31
Add colspan attribute to improve table semantics

Saulis:
Connects to #441

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/509)

### DIFF
--- a/test/column-groups.html
+++ b/test/column-groups.html
@@ -213,6 +213,18 @@
           expect(cell_group0._flexChanged.callCount).to.eql(1);
         });
 
+        it('should set colSpan to spanned header cells', function() {
+          var cell = getHeaderCell(0, 1);
+
+          expect(cell.colSpan).to.eql(2);
+        });
+
+        it('should not set colSpan to non-spanned header cells', function() {
+          var cell = getHeaderCell(0, 0);
+
+          expect(cell.hasAttribute('colspan')).to.be.false;
+        });
+
         it('should have right amount of rows', function() {
           expect(headerRows).to.have.length(2);
         });
--- a/vaadin-grid-column.html
+++ b/vaadin-grid-column.html
@@ -137,7 +137,10 @@
         _width: String
       },
 
-      observers: ['_flexChanged(_flex)', '_widthChanged(_width)', '_frozenChanged(_childColumns, frozen)'],
+      observers: ['_childColumnsChanged(_childColumns)',
+                  '_flexChanged(_flex)',
+                  '_widthChanged(_width)',
+                  '_frozenChanged(_childColumns, frozen)'],
 
       listeners: {
         'property-changed': '_columnPropChanged'
@@ -159,6 +162,12 @@
         if (e.detail.path === 'lastFrozen') {
           this._lastFrozen = e.detail.value;
         }
+      },
+
+      _childColumnsChanged: function(childColumns) {
+        // explicitly firing 'colSpan' change for the header cells.
+        // if needed, can be decoupled from cell by firing a 'childColumns' instead.
+        this.fire('property-changed', {path: 'colSpan', value: childColumns.length});
       },
 
       _updateFlexAndWidth: function() {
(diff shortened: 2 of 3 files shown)


manolo:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-table-cell.html, line 63 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509#-KVTfMFYwat_1bXZDI3J:-KVTfMFYwat_1bXZDI3K:bvuhzsi) ([raw file](https://github.com/vaadin/vaadin-grid/blob/8e2f9ca800ddf3c10903fbd8d040c0416767d349/vaadin-grid-table-cell.html#L63)):*
> 2.0-column-group-colspan
I feel like we don't need to add any new observer or event in column.
If you move this to the _flexChanged block every thing should work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509)*

manolo:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-table-cell.html, line 63 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509#-KVTfMFYwat_1bXZDI3J:-KVTmsKSTcZWfpXqmwXd:b-4zu6gp) ([raw file](https://github.com/vaadin/vaadin-grid/blob/8e2f9ca800ddf3c10903fbd8d040c0416767d349/vaadin-grid-table-cell.html#L63)):*
> > 2.0-column-group-colspan
> I feel like we don't need to add any new observer or event in column.
> If you move this to the _flexChanged block every thing should work.
</details>
Well, after thinking a bit more with this, I feel like we shouldn't know anything here about _childColumns, it should be the columns element the one setting a property, so that the code here should be `this.colSpan = column.colSpan` or `_colSpan` it should be readonly since it's computed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509)*

manolo:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-table-cell.html, line 63 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509#-KVTfMFYwat_1bXZDI3J:-KVTrfkoQoahEQyBWw4l:b-o0nk1x) ([raw file](https://github.com/vaadin/vaadin-grid/blob/8e2f9ca800ddf3c10903fbd8d040c0416767d349/vaadin-grid-table-cell.html#L63)):*
> Well, after thinking a bit more with this, I feel like we shouldn't know anything here about _childColumns, it should be the columns element the one setting a property, so that the code here should be `this.colSpan = column.colSpan` or `_colSpan` it should be readonly since it's computed.
</details>
Second round of thinking, the implementation using _childColumns is simpler than having to declare a new property in the column. So leave this as it is. We could define column.[_]colspan property in the future if needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509)*

manolo:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509#-:-KVTsBOEFJVrMLaykwIP:bnfp4nl)*

manolo:

---

*[vaadin-grid-table-cell.html, line 63 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509#-KVTfMFYwat_1bXZDI3J:-KVTwdJ8rjZMUMsybHnX:b-1ufvtk) ([raw file](https://github.com/vaadin/vaadin-grid/blob/8e2f9ca800ddf3c10903fbd8d040c0416767d349/vaadin-grid-table-cell.html#L63)):*
> Second round of thinking, the implementation using _childColumns is simpler than having to declare a new property in the column. So leave this as it is. We could define column.[_]colspan property in the future if needed.
</details>
Third round, I don't see any case where this block is needed, why not remove it? It should be enought with the event notification to set the property. No tests fail if the block is removed. It only makes sense when the cell is reused and the column changed, but that never happens in header/footers

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/509)*

Issue #534, opened 2016-11-17
Theming API


jouni:
Design and implement the necessary mixins for the grid.

tomivirkki:
# Theming API proposal

The aim of the research was to come up with a minimal theming API for the `vaadin-grid` 2.0 that could still cover a lot of the common use-cases when it comes to implementing a custom design.

The proposal suggests an MVP approach; the theming API can be expanded on demand anytime but providing too rich set initially might be counter-productive in terms of long-term maintenance.

The prototype API was formed by trying out how the current `vaadin-grid` 2.0-pre2 version could be styled to match a specific custom design and then deriving any repetitive patterns/parts whose theming turned out to be unnecessarily verbose or parts that were impossible to theme without proper APIs and turning them into built-in style mixins.

Two additional custom designs were implemented with the initial API to see if it serves it's purpose/to find any lackings.

The source for the current theming API prototype is at [2.0-proto-styling](https://github.com/vaadin/vaadin-grid/tree/2.0-proto-styling) branch. The demos are at ../demo/styling.html


## Notes

- We should agree on the default look and feel of the `vaadin-grid` before proceeding with the implementation.
- For many cases where the proposed API isn't sufficient per se, a proper solution may be to define named wrapper elements in the column templates instead of expanding the theming API to suit a particular use-case. (See "Styling 3" demo in the prototype demos for example)
- Dependencies/related issues (in priority order):
  - [Map flex into CSS flex instead of just flex-grow](https://github.com/vaadin/vaadin-grid/issues/527)
  - All cells must have a cell-content wrapper even if they didn't have a template (prototype has an implementation, need a separate issue perhaps?)
  - [Fix style scope issues](https://github.com/vaadin/vaadin-grid/issues/538)
  - [Empty header/footer rows shouldn't be displayed](https://github.com/vaadin/vaadin-grid/issues/549)

## Open questions
- Should our mixins target the `<td>`/`<th>` elements or the "cell-content" wrappers (`<div>`) directly inside of them?
  - Targeting the wrappers (`<div>`) might be safer since they're sort of isolated units that don't affect the grid's flow in any way
[...]

PR #555, opened 2016-12-03
All cells should have a content wrapper

tomivirkki:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/555)

### DIFF
--- a/test/column-groups.html
+++ b/test/column-groups.html
@@ -139,11 +139,11 @@
       var grid, scroller, header, body, footer, headerRows, footerRows;
 
       function getFooterCellContent(row, col) {
-        return getFooterCell(row, col).textContent.trim();
+        return getCellContent(getFooterCell(row, col)).textContent.trim();
       }
 
       function getHeaderCellContent(row, col) {
-        return getHeaderCell(row, col).textContent.trim();
+        return getCellContent(getHeaderCell(row, col)).textContent.trim();
       }
 
       function getHeaderCell(row, col) {
@@ -179,7 +179,8 @@
           var cells = Polymer.dom(header).querySelectorAll('th');
           expect(cells).not.to.be.empty;
 
-          cells.forEach(function(el) {
+          cells.forEach(function(cell) {
+            var el = getCellContent(cell);
             expect(el.matches(':empty')).to.be.true;
           });
         });
@@ -187,15 +188,16 @@
         it('should not have a border below the header', function() {
           var cells = Polymer.dom(header).querySelectorAll('th');
           var cellOnLastRow = cells[cells.length - 1];
-
-          expect(window.getComputedStyle(cellOnLastRow).borderBottomStyle).to.eql('none');
+          var el = getCellContent(cellOnLastRow);
+          expect(window.getComputedStyle(el).borderBottomStyle).to.eql('none');
         });
 
         it('should have empty footer', function() {
           var cells = Polymer.dom(footer).querySelectorAll('td');
           expect(cells).not.to.be.empty;
 
-          cells.forEach(function(el) {
+          cells.forEach(function(cell) {
+            var el = getCellContent(cell);
             expect(el.matches(':empty')).to.be.true;
           });
         });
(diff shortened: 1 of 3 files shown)


Saulis:
So what's the background for this? Why do all cells need to have a content wrapper?

tomivirkki:
It's needed for themability. Since we're (most likely) going to target the "cell-content" wrappers with our mixins, placing let's say a backgroud color for --vaadin-grid-header-cell mixin would result in empty squares in your header if every cell didn't contain a wrapper

Saulis:
Oh right.. should reconsider applying the mixins to cells instead? I remember you and Jouni thinking if it's a bit hacky to apply styles to generated items? Also, exposing (internal) class names to light DOM will prevent us from choosing freely any class names we want due to the risk of styles leaking from outside. (this is probably not a huge issue, but still..)

I think I would prefer keeping it all internal – in case the user applies some styles to the cells that conflict with the cell content wrappers, some general styles can still be applied from outside.

tomivirkki:
What we referred to as "hacky" with Jouni was enabling the user to directly apply CSS rules for the light DOM cell-content wrapper elements by classnames applied on the fly. This is not the case here. What we're planning is targeting the wrappers with the style mixins instead. It's actually more risky exposing mixins for the internal cells. Please read the "open questions" section at https://github.com/vaadin/vaadin-grid/issues/534#issuecomment-264159208

"in case the user applies some styles to the cells that conflict with the cell content wrappers, some general styles can still be applied from outside"
Again, we're really not considering adding classnames to the wrappers on the fly and you really shouldn't target them directly with any CSS rules. Their sole purpose is to provide an insertion point for arbitrary cell content so we can distribute light DOM nodes to grid cells in a clean manner.

Saulis:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-column.html, line 20 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/555#-KYNz-RkE0UR8zzAlwXK:-KYNz-RkEP_IBcSJJmy-:b-q558ii) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9edf3772d0fbfa066daf8bcdae55e4ff9f199cc5/vaadin-grid-column.html#L20)):*
> ```HTML
>           type: Object,
>           value: function() {
>             return this._findTemplate('template.header') || null;
> ```

Would it make more sense to add `document.createElement('template')` here as a default value instead of `null` ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/555)*

tomivirkki:
*[vaadin-grid-column.html, line 20 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/555#-KYNz-RkE0UR8zzAlwXK:-KYO7qCG0oNxCNNFO3si:b5i750y) ([raw file](https://github.com/vaadin/vaadin-grid/blob/9edf3772d0fbfa066daf8bcdae55e4ff9f199cc5/vaadin-grid-column.html#L20)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

Would it make more sense to add `document.createElement('template')` here as a default value instead of `null` ?
</blockquote></details>

This is a column-specific property (`headerTemplate`) but we need a template for each cell on that column (even the empty ones). This 'null' value is set here only to make `_headerTemplateChanged` invoke for cells on a column that don't have a `headerTemplate`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/555)*

Saulis:
:lgtm:

---

Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/555#-:-KYOnKps8dCYC3ChHjvO:bnfp4nl)*

PR #580, opened 2016-12-16
Templatize templates in separate templatizer elements

### LINKED ISSUE #531: Template bindings have limitations when using dom-repeat to define columns
Some limitation we discovered when implementing #503:

#### When using `dom-repeat`, some bindings only work on header/footer rows and 1st body row.
```html
<template is="dom-repeat" items="[[columns]]" as="column">
  <vaadin-grid-column>
    <template class="header">This works: [[column.foo]]</template>
    <template>
      <div>This only works for the 1st body row: [[column.foo]]</div>
    </template>
  </vaadin-grid-column>
</template>
```
In our tests, when `columns` property is *prepended* and `render()` is explicitly called on the `dom-repeat` element, the issue seems to (partly) disappear. Weirdly, when we *append* a new column, the issue remains.

...

Root cause of this issues seem to be the fact that we're calling `templatize()` multiple times for the body cell template.

### LINKED ISSUE #564: Row Details template doesn't have the correct context
The row details template should be applied the same `dataHost` as the `<vaadin-grid>` element has to allow using functions and event listeners declaratively.

Saulis:
Fixes #531 
Fixes #564 

This makes sure each template is only templatized once and that no element
never templatized more than one template.

Also, patches a memory leak which happened due to not removing template
instances from templateInstances array when cells get removed during column
tree update.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/580)

### DIFF
--- a/demo/columns.html
+++ b/demo/columns.html
@@ -12,6 +12,7 @@
 
   <link rel="import" href="../../paper-checkbox/paper-checkbox.html">
   <link rel="import" href="../../paper-input/paper-input.html">
+  <link rel="import" href="../../paper-button/paper-button.html">
   <link rel="import" href="../../iron-image/iron-image.html">
   <link rel="import" href="../../paper-menu/paper-menu.html">
   <link rel="import" href="../../paper-item/paper-item.html">
@@ -187,6 +188,43 @@
         </dom-module>
       </template>
       <x-grid></x-grid>
+    </demo-snippet>
+
+    <demo-snippet>
+      <h3>Dynamic Columns using dom-repeat</h3>
+      <template>
+        <x-dynamic-columns></x-dynamic-columns>
+        <dom-module id="x-dynamic-columns">
+          <template>
+            <x-data-source data-source="{{dataSource}}"></x-data-source>
+            <vaadin-grid id="grid" data-source="[[dataSource]]" size="200">
+              <template is="dom-repeat" items="[[columns]]" as="column">
+                <vaadin-grid-column>
+                  <template class="header">[[column]]</template>
+                  <template>[[_itemProperty(item, column)]]</template>
+                </vaadin-grid-column>
+              </template>
+            </vaadin-grid>
+          </template>
+          <script>
+            Polymer({
+              is: 'x-dynamic-columns',
+
+              properties: {
+                columns: Array
+              },
+
+              ready: function() {
+                this.columns = ['name.first', 'name.last', 'email'];
+              },
+
+              _itemProperty: function(item, column) {
+                return this.get('user.' + column, item);
+              }
+            });
+          </script>
+        </dom-module>
+      </template>
     </demo-snippet>
   </div>
 </body>
--- a/test/column-groups.html
+++ b/test/column-groups.html
@@ -187,8 +187,7 @@
           expect(cells).not.to.be.empty;
 
           cells.forEach(function(cell) {
-            var el = getCellContent(cell);
-            expect(el.matches(':empty')).to.be.true;
+            expect(cell._isEmpty).to.be.true;
           });
         });
 
@@ -204,8 +203,7 @@
           expect(cells).not.to.be.empty;
 
           cells.forEach(function(cell) {
-            var el = getCellContent(cell);
-            expect(el.matches(':empty')).to.be.true;
+            expect(cell._isEmpty).to.be.true;
           });
         });
 
--- a/vaadin-grid-column.html
+++ b/vaadin-grid-column.html
@@ -1,4 +1,5 @@
 <link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="vaadin-grid-templatizer.html">
 
 <dom-module id="vaadin-grid-column">
 </dom-module>
@@ -71,10 +72,22 @@
     },
 
     _headerTemplateChanged: function(headerTemplate) {
+      if (headerTemplate) {
+        var templatizer = new vaadin.elements.grid.Templatizer(this.dataHost);
+        templatizer._instanceProps = {};
+        templatizer.template = headerTemplate;
+      }
+
       this.fire('property-changed', {path: 'headerTemplate', value: headerTemplate});
     },
 
     _footerTemplateChanged: function(footerTemplate) {
+      if (footerTemplate) {
+        var templatizer = new vaadin.elements.grid.Templatizer(this.dataHost);
+        templatizer._instanceProps = {};
+        templatizer.template = footerTemplate;
+      }
+
       this.fire('property-changed', {path: 'footerTemplate', value: footerTemplate});
     },
 
@@ -125,6 +138,14 @@
     },
 
     _templateChanged: function(template) {
+      var templatizer = new vaadin.elements.grid.Templatizer(this.dataHost);
+
+      // body cell templatizer needs to be attached so that `item-changed` and
+      // `template-instance-changed` events propagate to grid.
+      Polymer.dom(this.root).appendChild(templatizer);
+
+      templatizer.template = template;
+
       // We bubble false for optimisation
       this.fire('property-changed', {path: 'template', value: template}, {bubbles: false});
     },
--- a/vaadin-grid-data-source-behavior.html
+++ b/vaadin-grid-data-source-behavior.html
@@ -65,6 +65,10 @@
    */
   vaadin.elements.grid.DataSourceBehavior = {
 
+    listeners: {
+      'item-changed': '_templateItemChanged'
+    },
+
     properties: {
 
       /**
@@ -101,6 +105,25 @@
           return {};
         }
       },
+    },
+
+    _templateItemChanged: function(e) {
+      var item = e.detail.item;
+
+      // TODO: We could avoid iterating rows by fixing _physicalIndexForKey mapping
+      // in iron-list-behavior so that vidx could be used to fetch the correct row element.
+      Polymer.dom(this.$.scroller.$.items).children.forEach(function(row) {
+        if (row.item === item) {
+          row.iterateCells(function(cell) {
+            // replacing the whole item instead of updating just the updated path
+            // to prevent _forwardInstancePath from firing duplicate `item-changed`
+            // events on all the sibling cells.
+            // main goal is just to get the [[item.xxx]] bindings up-to-date.
+            cell.instance.item = null;
+            cell.instance.item = item;
+          });
+        }
+      });
     },
 
     _getCachedItem: function(index) {
--- a/vaadin-grid-row-details-behavior.html
+++ b/vaadin-grid-row-details-behavior.html
@@ -1,3 +1,5 @@
+<link rel="import" href="vaadin-grid-templatizer.html">
+
 <dom-module id="vaadin-grid-row-details-styles">
   <template>
     <style>
@@ -39,7 +41,8 @@
     },
 
     observers: [
-      '_expandedItemsChanged(expandedItems.*, dataSource)'
+      '_expandedItemsChanged(expandedItems.*, dataSource)',
+      '_rowDetailsTemplateChanged(rowDetailsTemplate)'
     ],
 
     _expandedItemsChanged: function(expandedItems, dataSource) {
@@ -47,6 +50,21 @@
       this.$.scroller._physicalItems.forEach(function(row) {
         row.expanded = this._isExpanded(row.item);
       }.bind(this));
+    },
+
+    _rowDetailsTemplateChanged: function(rowDetailsTemplate) {
+      var templatizer = new vaadin.elements.grid.Templatizer(this.dataHost);
+      templatizer._instanceProps = {
+        expanded: true,
+        item: true,
+        selected: true
+      };
+
+      // row details templatizer needs to be attached so that `item-changed` and
+      // `template-instance-changed` events propagate to grid.
+      Polymer.dom(this.root).appendChild(templatizer);
+
+      templatizer.template = rowDetailsTemplate;
     },
 
     _isExpanded: function(item) {
--- a/vaadin-grid-table-cell.html
+++ b/vaadin-grid-table-cell.html
@@ -40,11 +40,11 @@
         target: Object,
         width: String,
 
-        _forwardedParentProps: Object,
         _childColumns: Array,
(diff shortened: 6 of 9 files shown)


manolo:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-grid-templatizer.html, line 6 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLgrjcrrx-FAahVYzf:-KZLgrjd11UTiRi_ZUTk:bxkdcxm) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-templatizer.html#L6)):*
> ```HTML
>   window.vaadin = window.vaadin || {};
>   vaadin.elements = vaadin.elements || {};
>   vaadin.elements.grid = vaadin.elements.grid || {};
> ```

I feel that we are repeating this same pattern in many places.
But these behaviours do not make sense without the grid so probably at this point we could assume that vaadin.element.grid should be declared

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

platosha:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-grid-templatizer.html, line 6 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLgrjcrrx-FAahVYzf:-KZLp1_ewXfIVV1GhIg-:b-1r56iq) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-templatizer.html#L6)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary><blockquote>

I feel that we are repeating this same pattern in many places.
But these behaviours do not make sense without the grid so probably at this point we could assume that vaadin.element.grid should be declared


</blockquote></details>

BTW It is the safest way. We can not guarantee and must not rely on the order of execution for these HTML imports, except for the requirements of each file. Especially in case of using processing tools like vulcanize, the order of execution for these imports might get unexpected.

This file’s requirements (i. e., polymer/polymer.html) do not declare this global scope, therefore we can not rely on its presence here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

tomivirkki:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-grid-selection-behavior.html, line 44 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLjoElOoi_XrRI98C1:-KZLjoElOoi_XrRI98C2:bj5jjea) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-selection-behavior.html#L44)):*
> ```HTML
> 
>     _isSelected: function(item) {
>       return item && this.selectedItems.indexOf(item) > -1;
> ```

Should check `...item !== null &&...` so falsy items (0) won't get ignored

---

*[vaadin-grid-table-cell.html, line 61 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZM_PJ336HhVHhxsCy9:-KZM_PJ336HhVHhxsCyA:bdt05ik) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-table-cell.html#L61)):*
> ```HTML
>       detached: function() {
>         if (this.template && this.template.templatizer) {
>           this.template.templatizer.removeInstance(this.instance);
> ```

The rows are reordered when scrolling stops. This will cause the relocated rows to detach/reattach -> all instances get removed from the templatizer eventually.

This needs to be async/debounced...same kind of an approach is already in use at `_toggleContent` (for the same reason)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

manolo:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-grid-table-cell.html, line 61 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZM_PJ336HhVHhxsCy9:-KZQdnPO-dg8dnf1yY9I:bclsgb4) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-table-cell.html#L61)):*

The rows are reordered when scrolling stops. This will cause the relocated rows to detach/reattach -> all instances get removed from the templatizer eventually.

This needs to be async/debounced...same kind of an approach is already in use at `_toggleContent` (for the same reason)
</blockquote></details>

Could be this done in a different way, so as the detach reattach does not implies template reconfiguration?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

manolo:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-grid-selection-behavior.html, line 44 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLjoElOoi_XrRI98C1:-KZQsZg9NInc-ta6ChsR:bajte4p) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-selection-behavior.html#L44)):*

Should check `...item !== null &&...` so falsy items (0) won't get ignored
</blockquote></details>

dont follow, what is the case? 

---

*[vaadin-grid-templatizer.html, line 6 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLgrjcrrx-FAahVYzf:-KZQgpqBZQ6jpf-PBDf6:b-d1820x) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-templatizer.html#L6)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

BTW It is the safest way. We can not guarantee and must not rely on the order of execution for these HTML imports, except for the requirements of each file. Especially in case of using processing tools like vulcanize, the order of execution for these imports might get unexpected.

This file’s requirements (i. e., polymer/polymer.html) do not declare this global scope, therefore we can not rely on its presence here.
</blockquote></details>

Sure, it's the safest way.
What I meant is that it's a very repetitive pattern, and the fact that if you debug for a while, when the block is executed those vars are always defined.
I don't think vulcanize needs to execute this code block for doing its work. Otherwise analyzer might do.
An option is to have some common import for doing this, but only if we have some other common stuff to share
Anyway, not a blocker. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

tomivirkki:
*[vaadin-grid-selection-behavior.html, line 44 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLjoElOoi_XrRI98C1:-KZQt9y-Xmz2DKwzO4au:bqlwkhd) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-selection-behavior.html#L44)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary><blockquote>

dont follow, what is the case? 
</blockquote></details>

For example `<vaadin-grid items="[0,1,2,3,4]">` doesn't work correctly with this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

manolo:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-grid-selection-behavior.html, line 44 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLjoElOoi_XrRI98C1:-KZQtl4FHyzOIpfdRA2c:byssnjn) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-selection-behavior.html#L44)):*

For example `<vaadin-grid items="[0,1,2,3,4]">` doesn't work correctly with this
</blockquote></details>

I see. So yes it needs to be checked against null and probably undefined. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

Saulis:
Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-cell.html, line 61 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZM_PJ336HhVHhxsCy9:-KZRYidC6pr3KjlV-ank:b-a14vyb) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-table-cell.html#L61)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary><blockquote>

Could be this done in a different way, so as the detach reattach does not implies template reconfiguration?
</blockquote></details>

Added a separate `attach` call. Manolo: the templatizer needs to propagate parent property changes to each template instance, so it needs to keep track of them. When a cell is discarded, we need to remove the instance from memory, otherwise it will keep a detached DOM tree in memory.

---

*[vaadin-grid-selection-behavior.html, line 44 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZLjoElOoi_XrRI98C1:-KZRIdIbEmWZUEhXWTWv:b-eyu9g) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-selection-behavior.html#L44)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary><blockquote>

I see. So yes it needs to be checked against null and probably undefined. 
</blockquote></details>

A better idea: I'll just remove the check (doesn't seem to be needed anymore :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

tomivirkki:
*[vaadin-grid-table-cell.html, line 61 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZM_PJ336HhVHhxsCy9:-KZRchpdhhLWaqg7MyUe:b-r0nbsn) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-table-cell.html#L61)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

Added a separate `attach` call. Manolo: the templatizer needs to propagate parent property changes to each template instance, so it needs to keep track of them. When a cell is discarded, we need to remove the instance from memory, otherwise it will keep a detached DOM tree in memory.
</blockquote></details>

The problem still exists as footer/header cells get deattached when the grid itself is deattached (for example relocated in the DOM or removed temporarily)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

tomivirkki:
*[vaadin-grid-data-source-behavior.html, line 117 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZVFkHHKTg0ArXSLn1W:-KZVFkHHKTg0ArXSLn1X:b290uuv) ([raw file](https://github.com/vaadin/vaadin-grid/blob/e22bddb591f8e4bdff90062c23be44af46aa2ebb/vaadin-grid-data-source-behavior.html#L117)):*
> ```HTML
>       Polymer.dom(this.$.scroller.$.items).children.forEach(function(row) {
>         if (row.item === item) {
>           Polymer.dom(row).children.forEach(function(cell) {
> ```

Same as `row.iterateCells(function(cell) {` ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

Saulis:
Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid-table-cell.html, line 61 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZM_PJ336HhVHhxsCy9:-KZVeyuJMGXd15eOxkpF:b-896fix) ([raw file](https://github.com/vaadin/vaadin-grid/blob/45b9965969e66c7c1cbb5092d230813b28d503ea/vaadin-grid-table-cell.html#L61)):*

The problem still exists as footer/header cells get deattached when the grid itself is deattached (for example relocated in the DOM or removed temporarily)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

Saulis:
Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-grid-data-source-behavior.html, line 117 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-KZVFkHHKTg0ArXSLn1W:-KZVhoGdu4e4pSVx7I-G:b-duzahh) ([raw file](https://github.com/vaadin/vaadin-grid/blob/e22bddb591f8e4bdff90062c23be44af46aa2ebb/vaadin-grid-data-source-behavior.html#L117)):*

Same as `row.iterateCells(function(cell) {` ?
</blockquote></details>

Yes! Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580)*

tomivirkki:
:lgtm:

---

Reviewed 5 of 10 files at r1, 2 of 4 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/580#-:-KZVxDZrJayUl2u70zjO:bnfp4nl)*

PR #585, opened 2016-12-19
Clear cache when data source changes.

charkins:
If the data source is changed, the grid will display incorrect data.
A similar observer was removed in 96db6c2 to address an issue with
the first page loading twice on startup, noting that there was
another observer in place for dataSource. There no longer appears
to be any other observer on the dataSource property, nor does this
change appear to cause a regression where the first page loads twice
on startup.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/585)

### DIFF
--- a/vaadin-grid-data-source-behavior.html
+++ b/vaadin-grid-data-source-behavior.html
@@ -83,7 +83,10 @@
        *
        * `params.pageSize` Current page size
        */
-      dataSource: Function,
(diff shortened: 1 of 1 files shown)


Saulis:
Awesome thanks! We really appreciate this!

(I'll run some checks and add a test before merging asap)

PR #640, opened 2017-01-04
2.0 dev scrollbars

depeele:
Update `vaadin-grid-table-scroll-behaviour:_onScroll` handler to determine if the scroll target has an ancestor prior to handling the scroll event directly.

Add scrollbar styling mixins to `vaadin-grid-table-themability-styles` in `vaadin-grid-table.html`.
Scrollbar styling is at this top-most level so it properly applies to the computed `scrollbarWidth` as well as the functional scrollbar that will appear on the raw table element.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/640)

### DIFF
--- a/vaadin-grid-table-scroll-behavior.html
+++ b/vaadin-grid-table-scroll-behavior.html
@@ -121,6 +121,26 @@
 
     _onWheel: function(e) {
       var table = this.$.table;
+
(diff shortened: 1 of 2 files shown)


depeele:
Closed to fix CI issues.

PR #641, opened 2017-01-04
Allow contained elements to be independently scrollable

depeele:
Update `_onWheel()` to  determine if the scroll target has an ancestor prior to the `vaadin-grid-table-scroll-behavior`  component that handles scrolling. In this case, do not handle the scroll event directly allowing it to continue to that ancestor.

This allows, for example, row-detail to contain a component that is scrollable. In this case, while the scroll target is that component, `vaadin-grid-table-scroll-behavior` will pass on handling the scroll event, allowing the event to be handles by that component.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/641)

### DIFF
--- a/vaadin-grid-table-scroll-behavior.html
+++ b/vaadin-grid-table-scroll-behavior.html
@@ -121,6 +121,26 @@
 
     _onWheel: function(e) {
       var table = this.$.table;
+
+      /* Determine if the scroll target has an ancestor prior to this
+       * table/scroller that handles scrolling
+       */
+      var el = e.target;
+      while (el && el !== table && el.scrollHeight === el.clientHeight) {
+        /* This element has a vertical scrollbar and so should receive and
+         * handle vertical scroll events.
+         */
+        el = el.parentElement;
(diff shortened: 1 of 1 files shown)


tomivirkki:
Good catch, thanks for the PR! this is definitely something that should be fixed before the 2.0 release.

Some issues regarding the PR:
 - Needs unit tests (to verify the issues and that they get fixed)
 - Seems to jam wheel scrolling with Safari (tested with Safari 10)
 - Doesn't have effect on IE11 / Edge

Let us know if you need any assistance with the PR

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/641#-:-Ka6Yj1b4BOL-gC9sOSr:blwyw9f)*

tomivirkki:
Hi, decided to extract this into a separate PR with the required fixes at #687. This PR will close once #687 gets merged.

Saulis:
Closed by #687

PR #642, opened 2017-01-04
Add scrollbar mixins for webkit-based browsers.

depeele:
Scrollbar styling is added in the `vaadin-grid-table-themability-styles` shared styles at the top-most level so it properly applies to the computed `scrollbarWidth` as well as the functional scrollbar that will appear on the raw table element.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/642)

### DIFF
--- a/vaadin-grid-table.html
+++ b/vaadin-grid-table.html
@@ -136,6 +136,25 @@
   <template>
     <style>
 
+      /* Scrollbars
+       *  @note Scrollbar styling is here at the top-most level so it properly
+       *        applies to the computed `scrollbarWidth` as well as the
+       *        functional scrollbar that will appear on the raw table element.
+       */
+
+      ::-webkit-scrollbar {
+        @apply(--scrollbar-common);
+      }
+
+      ::-webkit-scrollbar-thumb {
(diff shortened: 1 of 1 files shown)


tomivirkki:
Hi @depeele,

We discussed adding similar mixins during the themability API design phase and in the end decided not to include them in the default set. The reasons for leaving them out were mainly that the use-case is quite uncommon and most of all that the mixins are webkit-specific.

However, we have discussed adding a mechanism for including custom styles for the grid internals by adding a dedicated style module for the purpose. This would serve as a non-standard means for adding special styles that aren't covered by the built-in set of mixins.

Something like the following:
```
<dom-module id="foo-styles">
  <template>
    <style>
      ::-webkit-scrollbar {
        background: red;
      }
    </style>
  </template>
</dom-module>

<vaadin-grid style-module-name="foo-styles">
  ...
</vaadin-grid>
```

Note that this feature is not currently on our roadmap though.

...but at least for now, you should be able to target the internals with /deep/ selector in order to apply custom styles for the scrollbars.

Closing this PR for now

PR #725, opened 2017-02-21
Add keyboard access and a11y for sorting

### LINKED ISSUE #685: Sorting helpers cannot be accessed with keyboard
- Have also the screenreader announce something nice by default: "Hit x to sort desc/ascending"

platosha:
Fixes #685

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/725)

### DIFF
--- a/test/keyboard-navigation.html
+++ b/test/keyboard-navigation.html
@@ -1548,6 +1548,58 @@
             });
           }
         });
+
+        describe('focus events on cell content', function() {
+          var cellContent;
+
+          beforeEach(function() {
+            cellContent = getCell(0, 0)._cellContent;
+          });
+
+          it('should dispatch cell-focusin after trap focus', function() {
+            var spy = sinon.spy();
+            cellContent.addEventListener('cell-focusin', spy);
+
+            bodyTrap.focus();
+
+            expect(spy.callCount).to.equal(1);
+          });
+
+          it('should dispatch cell-focusout after trap blur', function() {
+            bodyTrap.focus();
+
+            var spy = sinon.spy();
+            cellContent.addEventListener('cell-focusout', spy);
+
+            headerTrap.focus();
+
+            expect(spy.callCount).to.equal(1);
+          });
+
+          it('should dispatch cell-focusin on keyboard navigation', function() {
+            bodyTrap.focus();
+            getCell(0, 1).click();
+
+            var spy = sinon.spy();
+            cellContent.addEventListener('cell-focusin', spy);
+
+            left();
+
+            expect(spy.callCount).to.equal(1);
+          });
+
+          it('should dispatch cell-focusout on keyboard navigation', function() {
+            bodyTrap.focus();
+            getCell(0, 0).click();
+
+            var spy = sinon.spy();
+            cellContent.addEventListener('cell-focusout', spy);
+
+            right();
+
+            expect(spy.callCount).to.equal(1);
+          });
+        });
       });
     }
   </script>
--- a/test/keyboard-navigation.html
+++ b/test/keyboard-navigation.html
@@ -47,6 +47,13 @@
           </template>
         </vaadin-grid-column>
         <vaadin-grid-column>
+          <template class="header">
+            <vaadin-grid-sorter path=""></vaadin-grid-sorter>
+          </template>
+          <template></template>
+          <template class="footer"></template>
+        </vaadin-grid-column>
+        <vaadin-grid-column>
           <template class="header"></template>
           <template>[[index]] [[item]]</template>
           <template class="footer"></template>
@@ -610,7 +617,7 @@
               expect(scroller._virtualFocus._focusedCellIndex).to.eql(0);
             });
 
-            it('should not navigate to hidden column with right arrow', function() {
+            it('should not navigate to hidden column with left arrow', function() {
               grid._columnTree[0][0].hidden = true;
               clickItem(0);
               left();
@@ -619,13 +626,14 @@
             });
 
             it('should not navigate to hidden column with right arrow', function() {
-              grid._columnTree[0][2].hidden = true;
+              grid._columnTree[0][3].hidden = true;
               clickItem(0);
 
               right();
               right();
+              right();
 
-              expect(scroller._virtualFocus._focusedCellIndex).to.eql(1);
+              expect(scroller._virtualFocus._focusedCellIndex).to.eql(2);
             });
 
             it('should not navigate to hidden column with end', function() {
@@ -638,12 +646,12 @@
             });
 
             it('should not navigate to hidden column with home', function() {
-              grid._columnTree[0][2].hidden = true;
+              grid._columnTree[0][3].hidden = true;
               clickItem(0);
 
               end();
 
-              expect(scroller._virtualFocus._focusedCellIndex).to.eql(1);
+              expect(scroller._virtualFocus._focusedCellIndex).to.eql(2);
             });
           });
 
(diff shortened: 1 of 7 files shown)


tomivirkki:
I was expecting accessing sorters to work the same as other fields placed in the grid cells. Meaning that you first enter the cell in interaction state (with enter) and then interact with the sorters (with space). This should be doable by giving the sorter a default tab-index and handling “space” key as a `_cellClick`.

Your approach is easier to use but it breaks the pattern and requires a lot of special handling for `<vaadin-grid-sorter>` in “vaadin-grid-cell-click-behacvior.html”, “vaadin-grid-focusable-cell-container-behavior.html” and “vaadin-grid-keyboard-navigation-behavior.html”

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-grid-table-focus-trap.html, line 16 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-KdUqx8RBXpq4Izx_zbh:-KdUqx8RBXpq4Izx_zbi:bn7eavx) ([raw file](https://github.com/vaadin/vaadin-grid/blob/8a5322bdfe111d6100c71659b4e9b27902d1fe64/vaadin-grid-table-focus-trap.html#L16)):*
> ```HTML
> 
>     
>     <div class="primary" tabindex="0" role="gridcell" on-focus="_onBaitFocus" on-blur="_onBaitBlur"><div style="position:absolute;z-index:1;" aria-hidden="true">&nbsp;</div></div>
> ```

Should probably say `_onTrapFocus` and `_onTrapBlur`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-KdUtqjOTIxmwr8-bMif:bdeh64m)*

platosha:
@tomivirkki Right. Entering interactive mode, though, wouldn't play nice when there is a filter element inside sorter.

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-grid-table-focus-trap.html, line 16 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-KdUqx8RBXpq4Izx_zbh:-KdUvocc9BlsNqrZCDxB:b-1yudes) ([raw file](https://github.com/vaadin/vaadin-grid/blob/8a5322bdfe111d6100c71659b4e9b27902d1fe64/vaadin-grid-table-focus-trap.html#L16)):*

Should probably say `_onTrapFocus` and `_onTrapBlur`?
</blockquote></details>

Well, the hosting element is named -trap, but these listeners are not for the host. I think, not mentioning trap here is less confusing in the end.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-KdUvBDBhfxTaA1KK0cx:bcb9fhz)*

tomivirkki:
You should be able to tab between cell fields while in interaction mode (in that case between the filter and the sorter, right?)

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-KdUxiQOc3ivFKwq8Rdg:b-xc9clg)*

manolo:
Quick Tested with both ways, space and Enter+space, and I feel a better ux using just the space.
I think sorting is a grid intrinsic feature when enabled, so it's reasonable that you have a shortcut for that. Probably this is something that could be decided by the PO or PD. Ping @jouni 

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-KdduZUSPKm9HI8LuVWM:bj30xjw)*

tomivirkki:
Yes it’s end-user friendlier, but it’s still a separate component from grid, you even need to import is explicitly in order to use it. IIRC the way the grid currently works with fields follows some accessibility spec so I wouldn’t add some special exception for this, at least now. The ticket was about making the sorter work with keyboard (like other fields do), there’s no mention about including special handling to improve UX in this case, that’s clearly scope creep.

We can later consider changing the grid’s behaviour so that hitting “space” on a cell automatically enters the interactive state and forwards the “space” as a click to an internal field but that should work the same way with all fields, not just one. Having this one random exception can also be unexpected for the end-user.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-Kddza_HmSuzcjvhnKhT:b82q224)*

manolo:
You win ;-), +1 to going with standards here and improve accessibility later in case we decide that.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-Kde4kXOEUZLK4tueA6E:bm766hf)*

tomivirkki:
Discussed this and decided to proceed with this PR as it is. But instead of making `<vaadin-grid-sorter>` the exception, all `<button>`s and elements with `role=“button”` should work like that (directly accessible on cell focus). Sorter will be decorated with `role=“button”` and given a default tab index. This approach will be better in line with “Example 2” at https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-KdeOERpN5KigIvGSDAg:bkkgqxl)*

jouni:
If I understood correctly, we will add a convenience feature, that if a cell contains a `role="button"` element, <kbd>Space</kbd> will trigger that in the navigation mode as well, when there’s only one such element inside the cell, right?

How does that play together with selecting a row? I suppose we should cancel selection then on those kinds of cells?

platosha:
Changing the keyboard navigation so that button-like elements are focused directly could be hard to implement. For a quicker fix for now, I would instead just add the convenience <kbd>Space</kbd>-click feature.

Not sure if it makes sense to apply “only one button-like” element restriction if there is more such elements inside a cell. Maybe let the space simply hit the first button? Pressing <kbd>Enter</kbd>, as usual, enters the interactive mode and focuses the first button, so that the other buttons are available to focus with <kbd>Tab</kbd> then.

Selecting a row indeed should be disabled for these cells then.

platosha:
Changed the space action to click on the first child element of a cell, if any.

In the cell click behaviour, there is an explicit announce call if the sorter was clicked. Removing that turned out to be hard with the current virtual focus implementation.

As discussed and quickly tried with @saulis, we should consider using native focus, it might make things simplier.

---

Review status: 0 of 33 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-Ke4PgGXiWfNxnDDJgNl:bo4wmw6)*

platosha:
Recent changes in brief:

- improved the a11y announce for the sorter,
- removed explicit sorter announce call on toggle from the grid,
- added an ARIA live zone in the sorter element instead.

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-KeDas0XU0zZulh3crm_:bxo1el5)*

Saulis:
:lgtm:

---

Reviewed 1 of 4 files at r2, 1 of 30 files at r3, 5 of 30 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/725#-:-KeE6c9-cEUIGCnTI7MF:bnfp4nl)*

PR #749, opened 2017-02-28
Change scroller sizing to avoid a Safari layouting issue.

tomivirkki:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/749)


tomivirkki:
The issue is quite visible at https://vaadin.github.io/expense-manager-demo/ (scroller is not as high as the grid itself). Safari only.

PR #774, opened 2017-03-06
Make multi-sorting opt-in

### LINKED ISSUE #742: Multi-column sorting should be opt-in
When using `<vaadin-grid-sorter>` elements, the default behavior should be that only one of them is “active” at any time, so that the grid is sorted by one column at a time.

Multi-column sorting should be an opt-in feature, either by the developer of by the end user.

In vaadin-grid 1.x, it’s opt-in by the end user (shift-click or long-tap), but that makes it really difficult to discover. I suggest we make it opt-in for the developer in 2.0, so that the developer needs to do some extra configuration to enable multi-column sorting, and then it will function like it does currently for the end user (normal clicks change the column sorting independently from other columns).

### API proposal
- By default, the `<vaadin-grid-sorter>` element overrides the sorters array with just one sorter element (which in result resets all other vaadin-grid-sorter elements)
- Add a new `independent` attribute for the `<vaadin-grid-sorter>` element. When that attribute is added, that sorter does not reset the whole sorters array, but instead prepends itself to the array.

Another option would be to introduce some attribute/property on the vaadin-grid level, that would affect all vaadin-grid-sorter elements simultaneously.

Saulis:
Fixes #742

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/774)

### DIFF
--- a/demo/sorting-filtering.html
+++ b/demo/sorting-filtering.html
@@ -13,12 +13,20 @@
   <link rel="import" href="../vaadin-grid-sorter.html">
   <link rel="import" href="../vaadin-grid-filter.html">
 
+  <link rel="import" href="../../paper-checkbox/paper-checkbox.html">
+
   <link rel="import" href="common.html">
 </head>
 
 <body unresolved>
   <div class="vertical-section-container centered">
     <demo-navigation></demo-navigation>
+
+    <style>
+      paper-checkbox {
+        margin-bottom: 10px;
+      }
+    </style>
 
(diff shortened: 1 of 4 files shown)


tomivirkki:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/774#-:-KeYwPyVMQdHYHMgZHY8:bnfp4nl)*

PR #802, opened 2017-03-13
Fix demos to work IE11

### LINKED ISSUE #799: Demos with custom dom modules don't work on IE11
Since IE11 uses a polyfill for `<template>`, any `<script>` inside a template is run during init.

Demo which include a template containing custom `<dom-module>` and `<script>` which calls `Polymer({..})` will fail in IE11.

Adding a `WebComponentsReady` listener should fix this issue.

platosha:
Fixes #799

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/802)


limonte:
This change violates the DRY principle. 

This is the **dev** documentation, a number of developers who using IE to read our docs is 0%, so we don't probably want to show the same useless code in each example for other 100% visitors?

platosha:
@limonte Can agree on DRY.

Mind that this is not quite useless code, though, since it is required to make the code work as expected in the main document in all browsers. This is kind of a documented requirement: https://www.polymer-project.org/1.0/docs/devguide/registering-elements#main-document-definitions If we register elements synchronously in the main document, then we violate Polymer’s docs on the subject, right?

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/802#-:-Kf7AkjyweCi9NQgUK3d:be97a0u)*

manolo:
We already use this approach in other elements (grid 1.x, combo, upload, etc) in order to fix polyfilled browsers.
It only manifests in IE by chance because it takes more time on asynchronously register elements than FF for instance.
Actually if the user copy and paste the code outside the `demo-snippet` it happens on those browsers as well.

To me is OK to have this callbacks here because
1- We are coherent with solution given for IE11 other elements
2- Copy and Paste is guaranteed to work
2- Polymer users already are familiar with WebComponentsReady callback

Unless @limonte has a better approach that could be used in all elements :lgtm:

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/802#-:-KfCUM9L4AJYVYXEaW-_:b-vypjab)*

limonte:
In our examples we are not trying to achieve cross-browser compatibility. We are showing isolated pieces of code to demonstrate different parts of component's functionality.

I definitely think that this should be handled in demo-snippet component, not in every single example. 
Here's my try: https://github.com/PolymerElements/iron-demo-helpers/pull/46 Unfortunately it was not good enough to be merged.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/802#-:-KfL02fWDLD4cF22cZqH:b-qgok93)*

manolo:
Not sure about this should be fixed in the demo snippet. Delaying everything in that component is not optimal. Think that you can wrap in the component any html/javascript code not only web-components or polymer related ones, and you must not penalise them with this delay.
Monica also recommends to handle this case in the demo.
Hence we need this fix anyways. 

:lgtm_strong: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/802#-:-KfL2F3iVeAiXz9RFvJ0:b-tcd2zx)*

limonte:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/802#-:-KfLN7YmUL5BX_d2XxCC:bnfp4nl)*

PR #826, opened 2017-03-28
Doc: Add default value

stramel:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/826)

### DIFF
--- a/vaadin-grid-column.html
+++ b/vaadin-grid-column.html
@@ -37,6 +37,7 @@ The `class` attribute is used to differentiate header and footer templates from 
 
       /**
        * When set to true, the column is user-resizable.
+       * @default false
        */
       resizable: {
         type: Boolean,


Saulis:
Hi!

Isn't the default value is currently `undefined` ? Should we make it `false` at the same time what do you think?

stramel:
@Saulis Currently, the docs are displaying `UNKNOWN` since it is a function. Adding `@default` should allow you to set it. You are correct it implicitly returns `undefined` currently on a `Boolean` property. I have no problem either way, just assumed Polymer would treat `undefined` as false on a `Boolean` property. Thinking about it more, it probably leaves it as `undefined`.  JSFiddle confirms it. https://jsfiddle.net/stramel/xpv5u9cc/2/

Would you like me to change the documentation to `undefined` or default property to `false` instead of `undefined`?

Saulis:
@stramel I think making it would be an improvement to have both the documentation and actual default value to be `false` – if you could add an else clause returning `false` I'd be happy to merge this PR. Thanks!

Saulis:
@stramel actually, nevermind – I'll give you a hand and add the else clause :smile:. 
Thanks for the PR!

PR #833, opened 2017-03-29
Set virtual focus to null async to work with async focus handling in IE11

### LINKED ISSUE #832: IE11: can't tab out from grid when there is no footer

Saulis:
Fixes #832 

There a caveat with this fix: in IE11 user still needs to tab through the footer even if it's not visible. To fix this, the whole focus trapping system needs a rewrite, basically removing the body trap and having the focus in the footer trap while user is navigating either in the body or footer. 

Now, this requires some work and the focus trapping system has been rewritten in the branch for Polymer 2 support, so I'm leaning towards living with this caveat for now.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/833)

### DIFF
--- a/test/keyboard-navigation.html
+++ b/test/keyboard-navigation.html
@@ -449,17 +449,21 @@
             expect(bodyTrap.focus.callCount).to.eql(1);
           });
 
-          it('should not move virtual focus from body to footer on tab when no footer rows exist', function() {
+          it('should not move virtual focus from body to footer on tab when no footer rows exist', function(done) {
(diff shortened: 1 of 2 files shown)


tomivirkki:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/833#-:-KgOpTVOAETkD9YiLxA0:bnfp4nl)*

PR #856, opened 2017-04-21
Improve performance during vaadin-grid.clearCache for grids with row details

craPkit:
When clearing the cache of a grid with templates, all rows were marked as expanded (due to a null item being passed to _isExpanded). Therefore the row-detail template was stamped for each row, just for teardown purposes. Performance used to be particularly bad, naturally, for complex templates * many rows.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/856)

### DIFF
--- a/vaadin-grid-row-details-behavior.html
+++ b/vaadin-grid-row-details-behavior.html
@@ -81,7 +81,7 @@
     },
 
     _isExpanded: function(item) {
-      return this.expandedItems && this.expandedItems.indexOf(item) !== -1;
+      return !!item && this.expandedItems && this.expandedItems.indexOf(item) !== -1;
     },
 
     /**
--- a/vaadin-grid-row-details-behavior.html
+++ b/vaadin-grid-row-details-behavior.html
@@ -15,7 +15,7 @@
 <dom-module id="vaadin-grid-row-details-themability-styles">
   <template>
     <style>
-      td[detailscell] ::content > vaadin-grid-cell-content {
+      td[detailscell] ::content>vaadin-grid-cell-content {
         background: #fff;
         @apply(--vaadin-grid-body-row-details-cell);
       }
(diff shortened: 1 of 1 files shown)


Saulis:
Hi there @craPkit!

The problem sounds pretty bad but I'm not sure how to reproduce it? Also, I don't quite understand how the proposed fix will fix it – since `this.expandedItems.indexOf(null)` will always return `-1`. (unless `this.expandedItems` contains `null` values of course)

Can you provide a test which shows the issue?

Thanks!

craPkit:
Hi Sauli and thanks for your reply!

It might take some time for me to provide a test case.
In my case there obviously exists a null value in expandedItems. I will try
to look into the actual cause next week, but I suppose it might have to do
with me using a data provider function.
I have a rather complicated setup since I'm constantly hot-swapping data
sources (i.e. different components, each with dynamically created requests
and partial data fetching using indices) in my application...

Cheers,
Patrick

manolo:
Shouldn't we avoid instead that expanded items never contains a null in `expandItem`?  is there a good reason for having a null item in the array ?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/856#-:-KkZa0NcAdbpoAFR0PLP:b-t4rk63)*

stramel:
I agree, it sounds like there shouldn't ever be a `null` in the `expandedItems` but does it make sense to also not even try to check the index if `null` is passed into the function?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/856#-:-KmCfIgFhToT3OQRu7cf:bilf7pj)*

tomivirkki:
Closing as inactive

PR #891, opened 2017-06-05
Update vaadin-grid-array-data-provider-behavior.html

leebickmtu:
Fix _checkPaths function.

The intention of line 75 is to remove the last level of the path. a.b.c -> a.b
Currently the regex matches and replaces the first occurrence of .X resulting in a.b.c -> a.c

Fixed by changing the regex to last last occurrence.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/891)

### DIFF
--- a/vaadin-grid-array-data-provider-behavior.html
+++ b/vaadin-grid-array-data-provider-behavior.html
@@ -72,7 +72,7 @@
           continue;
         }
 
-        var parentProperty = path.replace(/\.[^\.]*/, ''); // a.b.c -> a.b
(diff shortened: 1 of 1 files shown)


tomivirkki:
Good find, thanks :lgtm: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/891#-:-KlrQnX4Vp587_6Eulw2:bimjdns)*

leebickmtu:
Please also patch onto the 2.0.2 release as well, that is where I found the issue and need the fix.

PR #911, opened 2017-06-22
Creates items array if is undefined in crud demo

obedsoto:
Make sure items array exists before the manipulation

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/911)

### DIFF
--- a/demo/crud.html
+++ b/demo/crud.html
@@ -141,6 +141,7 @@
 
                 _add: function(e) {
                   if (this.$.firstname.value !== '' && this.$.lastname.value !== '') {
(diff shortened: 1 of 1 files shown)


obedsoto:
CLA signed.

Saulis:
Hi! Can you show an example in which case the `items` would be `undefined` ? It should be always set by the `x-array-data-source` before any `_add` or `_remove` functions are called.

obedsoto:
Hi @Saulis, yes you are right, but in my case I used this example with no data, I am using the element to let the user to add records, so there's no initial data to append. Demo works great with a data source.

tomivirkki:
Closing as wontfix

PR #916, opened 2017-06-27
Fix bug with focus trap extra spacing

### LINKED ISSUE #915: vaadin-grid-table-focus-trap adds extra space to the grid element
vaadin-grid-table-focus-trap adds extra space to the bottom of the grid. It has height 0px, but inside of it there are 2 divs (secondary and primary) that overflow this element and setting height of the grid doesn’t work properly (real height of the grid is more than I’m trying to set because of this 2 elements).

Toshegg:
Resolves #915

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/916)

### DIFF
--- a/test/scrolling-mode.html
+++ b/test/scrolling-mode.html
@@ -133,6 +133,10 @@
         expect(tableWheel).to.not.throw(Error);
       });
 
+      it('should have proper height', function() {
+        expect(grid.scrollHeight - grid.clientHeight).to.equal(0);
+      });
+
     });
   </script>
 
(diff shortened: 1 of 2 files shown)


tomivirkki:
All bugfixes should have a test case also to:
1. Verify the original bug
2. Make sure the bug doesn't accidentally re-emerge

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/916#-:-KnhZ64sOgt9_atP22o6:bdwm56r)*

PR #936, opened 2017-07-27
Fix removing not existing instance.

Yavanosta:
Splice with -1 as first argument will remove last item from array.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/936)


tomivirkki:
Thanks for the PR, @Yavanosta. In what circumstance can the argument of `removeInstance` be missing from the `_templateInstances` array? Can you provide a test case with this PR that proves the error?

tomivirkki:
Closing as inactive

Issue #878, opened 2017-05-17
Convert to ES6


platosha:
re enable iphone ipad tests

tomivirkki:
Known issues that need to be taken care of in scope of this:
- [ ] `<iron-list>` is currently inherited by the `<vaadin-grid>`. Mark all iron-list public API's as `@private` (but do it in vaadin-grid-scroller) so they don't end up in the apidoc
- [ ] `scrolling` attribute should not be applied to host. Move it to the `#scroller` element instead and rescope selectors
- [ ] Scope `VaadinGridScroller` under Vaadin.Grid like the rest of the modules
- [ ] Column resizing needs to mind affecting padding/border etc of `<th>` elements also when computing minimum width for a column
- [ ] Mobile Chrome doesn't allow horizontal touch scrolling over the header
- [ ] While reordering column groups, not all cells have the correct indicator colors
- [ ] Grabbing a column resizer handle can activate column reordering
- [ ] Fix layout issues in IE11

PR #961, opened 2017-09-04
Fix(keyboard): resizing focuses header cell

platosha:
Connected to #878

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/961)

### DIFF
--- a/test/column-resizing.html
+++ b/test/column-resizing.html
@@ -179,6 +179,18 @@
         expect(spy.called).to.be.true;
       });
 
+      it('should prevent default mousedown action on resize handle', function() {
+        var event = new CustomEvent('mousedown', {bubbles: true, cancelable: true, composed: true});
+        handle.dispatchEvent(event);
+        expect(event.defaultPrevented).to.be.true;
+      });
+
+      it('should allow default mousedown action on cells', function() {
+        var event = new CustomEvent('mousedown', {bubbles: true, cancelable: true, composed: true});
+        headerCells[0].dispatchEvent(event);
+        expect(event.defaultPrevented).to.be.false;
+      });
+
     });
 
     describe('column group resizing', function() {
--- a/vaadin-grid-column-resizing-mixin.html
+++ b/vaadin-grid-column-resizing-mixin.html
@@ -61,6 +61,9 @@ This program is available under Apache License Version 2.0, available at https:/
 
       // Disable contextmenu on any resize separator.
       scroller.addEventListener('contextmenu', e => e.target.getAttribute('part') == 'resize-handle' && e.preventDefault());
+
+      // Disable native cell focus when resizing
+      scroller.addEventListener('mousedown', e => e.target.getAttribute('part') === 'resize-handle' && e.preventDefault());
     }
 
     _onHeaderTrack(e) {


manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-column-resizing-mixin.html, line 66 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961#-KtBV7CQFN4mDDkTIeox:-KtBV7CQFN4mDDkTIeoy:bnvfwsu) ([raw file](https://github.com/vaadin/vaadin-grid/blob/e0a3465b405306769195b8b00cba8064c0fd228b/vaadin-grid-column-resizing-mixin.html#L66)):*
> ```HTML
> 
>       // Disable native cell focus when resizing
>       scroller.addEventListener('mousedown', e => e.target.getAttribute('part') === 'resize-handle' && e.preventDefault());
> ```

nit: can you align equal operator with the line above. I'm not fan on the triple operator in this case, but anyone works to me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961)*

platosha:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-column-resizing-mixin.html, line 66 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961#-KtBV7CQFN4mDDkTIeox:-KtBYbKSXbYyIPcywCNw:byvbbfm) ([raw file](https://github.com/vaadin/vaadin-grid/blob/e0a3465b405306769195b8b00cba8064c0fd228b/vaadin-grid-column-resizing-mixin.html#L66)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary><blockquote>

nit: can you align equal operator with the line above. I'm not fan on the triple operator in this case, but anyone works to me.
</blockquote></details>

Not sure about this. The line above is the only double-equal operator in this file, all the others are triple-equal.

IMO, triple equal operator should be preferred and enforced with a linter rule, as it’s more strict. We had this at some point.

Seems to me like it’s rather the line above should be with everything else, rather then this new line by me. But that is a separate issue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961)*

tomivirkki:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-grid-column-resizing-mixin.html, line 66 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961#-KtBV7CQFN4mDDkTIeox:-KtD-GG6IOQhw1L4YYlN:b-v27xs5) ([raw file](https://github.com/vaadin/vaadin-grid/blob/e0a3465b405306769195b8b00cba8064c0fd228b/vaadin-grid-column-resizing-mixin.html#L66)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Not sure about this. The line above is the only double-equal operator in this file, all the others are triple-equal.

IMO, triple equal operator should be preferred and enforced with a linter rule, as it’s more strict. We had this at some point.

Seems to me like it’s rather the line above should be with everything else, rather then this new line by me. But that is a separate issue.
</blockquote></details>

"My advice is to never use the evil twins. Instead, always use === and !==" (Douglas Crockford)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961#-:-KtD-NJxtSrjw5Z4dw67:bnfp4nl)*

manolo:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[vaadin-grid-column-resizing-mixin.html, line 66 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961#-KtBV7CQFN4mDDkTIeox:-KtG1Mc-rZrd3gVvDF8O:bq648ng) ([raw file](https://github.com/vaadin/vaadin-grid/blob/e0a3465b405306769195b8b00cba8064c0fd228b/vaadin-grid-column-resizing-mixin.html#L66)):*

"My advice is to never use the evil twins. Instead, always use === and !==" (Douglas Crockford)
</blockquote></details>

Yep, it's one of the recurrent problems of dealing with JS. So his solution is just 'I follow my own convention' I forgot about the double operator to avoid problems. But double operators are easy to read for developers handling multiple languages, and actually the triple operator only makes full sense when checking zero nulls empties undefined, in this case where right side is a string constant it does not aport any advantage, but the convention. 

Anyways, I just wanted to have both lines aligned instead of starting a discussion thread. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/961#-:-KtG37iMNGEDUHnRSAPY:bnfp4nl)*

PR #1027, opened 2017-10-04
Never set template instance item to null or empty

### LINKED ISSUE #1024: [4.0.0-alpha3] Computed functions in vaadin-grid-column templates called with empty or null item objects
Not sure where exactly this issue resides, but if you have a computed function in a `vaadin-grid-column` template, it gets called way more often than you would think. Beyond just a performance issue, there is the problem that the `item` object can be empty or `null` for some reason during the calls. **It will throw JS errors out when you scroll the grid** as `null` objects are sent to the function. I have a jsbin that illustrates the point:

https://jsbin.com/libiyiy/edit?html,console,output

Run the JS, you will see empty objects, scroll to the bottom, you will see `null` objects and JS errors in the console.

tomivirkki:
Fixes #1024

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1027)

### DIFF
--- a/test/templates.html
+++ b/test/templates.html
@@ -20,7 +20,7 @@
     <template>
       <vaadin-grid size="10" id="grid" style="height: 300px" data-provider="[[dataProvider]]">
         <vaadin-grid-column>
-          <template>[[_format(index)]] [[parentProp]]</template>
+          <template>[[_format(index)]] [[parentProp]] [[_formatItem(item)]]</template>
           <template class="header">[[_format('header1')]] [[parentProp]]</template>
           <template class="footer">Footer</template>
         </vaadin-grid-column>
@@ -76,6 +76,9 @@
           }
           _format(value) {
             return 'foo' + value;
+          }
+          _formatItem(item) {
+            return item;
           }
           _valueChanged() {
             this.valueChanged = true;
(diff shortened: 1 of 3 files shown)


test/templates.html
```diff
@@ -162,6 +165,24 @@
...
+          grid._scrollToIndex(100);
+          spy.getCalls().forEach(call => {
+            expect(call.args[0]).not.to.be.null;
+          });
+        });
+
+        it('should not invoke computed functions with empty item', () => {
+          const spy = sinon.spy(container, '_formatItem');
+          grid.size = 1000;
+          grid._scrollToIndex(100);
```
moderndeveloperllc: You may want to set the index to be higher. I only got this error when scrolling towards the end of a long list of items. Unless you were able to get this to fail before you put in the fix, in that case ignore :-)

tomivirkki: Both tests are failing without the fix. It's enough to scroll past the currently cached page (50 first items in this case)

platosha:
:lgtm:

---

Reviewed 1 of 2 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1027#-:-Kvg2l2G_dEG0dhZ8G2K:bnfp4nl)*

PR #1064, opened 2017-10-27
Get filter/sort changed event emitter from path under shadow dom

n1ywb:
As a developer I want to instantiate a dynamic set of columns configured by data. I define one column element for each type of data I want to be able to display and use dom-repeat and dom-if to get the right column type for the right data type. I also want my column headers to also be dynamic wrt to turning on filtering, sorting, or both. And I want each one of my dynamic column types to have the same dynamic header. Therefor I want to encapsulate vaadin-grid-sorter and vaadin-grid-filter and the business logic to configure them inside my own custom element which I can then insert into the light dom of the column header templates. This works perfectly under shady DOM, but not under shadow dom, because of event retargeting. This change causes the event listeners to use the event path property in preference of the target property if available, which allows the listener to get a reference to the filter/sorter elements even when they are nested inside a wrapper element's shadow root.

unit tests are included.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1064)

### DIFF
--- a/test/filtering.html
+++ b/test/filtering.html
@@ -23,14 +23,21 @@
           display: block;
         }
       </style>
-      <vaadin-grid-filter path="foo" value="[[_filterValue]]" id="filter">
-        <input value="{{_filterValue::input}}">
+      <vaadin-grid-filter
+        path="foo"
+        value="[[filterValue]]"
+        id="filter"
+      >
+        <input value="{{filterValue::input}}">
       </vaadin-grid-filter>
     </template>
     <script>
       HTMLImports.whenReady(function() {
         Polymer({
-          is: 'filter-wrapper'
+          is: 'filter-wrapper',
+          properties: {
+            filterValue: String
+          }
         });
       });
     </script>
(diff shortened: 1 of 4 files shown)


tomivirkki:
LGTM

tomivirkki:
Oh, sorry, I'll need to take that LGTM back after doing some more thinking. So even though the approach works it has one fundamental flaw: Grid holds references to the sorter/filter elements and even applies modifications to them (the multi-sort order numbering). Now, if the filter is contained in another element's shadow root, the grid should have no direct visibility to it.

For the mentioned use-case I would suggest you to make your wrapper implement the same API that the actual contained sorter/filter element so that the grid would'n communicate with the filters directly but instead trough the wrapper's matching APIs. You'll probably also need to stop the internal "sorter/filter-changed" event on the wrapper level and fire a new event. This "workaround" requires some boiler-plate piping but the important thing is that it doesn't violate the shadow root boundaries.

tomivirkki:
Closing as inactive

PR #1065, opened 2017-10-27
add explicit native shadow dom run to test suite

n1ywb:
by default, WCT runs one test run using the default DOM model for that version of polymer on that browser

this means, among other things, that currently you are not running tests AT ALL against Polymer 1 native shadow dom on chrome.

I have corrected the test suite so it explicitly runs twice, once under shady, and once under shadow, if supported. This is the approach suggested by the Polymer documentation. https://www.polymer-project.org/1.0/docs/tools/tests#shadow-dom

I notice that a few tests fail under polymer 1 native shadow dom on chrome. These are most likely bugs in the tests and or the element WRT shadow dom.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1065)

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -17,7 +17,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 </head>
 <body>
   <script>
-    WCT.loadSuites([
+      let suites = [
       'basic.html',
       'app-localize-behavior.html',
       'column.html',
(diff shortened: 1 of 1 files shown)


tomivirkki:
We actually used to have this setup in the test suite but removed it to mitigate CI issues. Merging this would cause CI build to exceed the running limits on Travis. The tests can be run with shadow dom locally with `?dom=shadow`.

PR #1066, opened 2017-10-29
This fix is to trap null values of observer parameters causing the sorter-changed event to fire 2 times and to add details on sort changed event.

nexuslifeline:
I was actually trying to sort the entire list with pagination. Apparently, the built in sort behavior of the element is actually sorting those on current page only and not the entire list. And I believe, to be able to sort the entire list we must actually sort the data source of vaadin grid by handling it to the event sort-changed emmitted by it, and so, we need to include some details upon dispatching this event particularly the column sort path and the direction of sorting.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1066)

### DIFF
--- a/vaadin-grid-array-data-provider-behavior.html
+++ b/vaadin-grid-array-data-provider-behavior.html
@@ -86,11 +86,14 @@
     },
 
     _multiSort: function(a, b) {
+        var sa = a.toString().toLowerCase();
+        var sb = b.toString().toLowerCase();
+
       return this._sorters.map(function(sort) {
         if (sort.direction === 'asc') {
-          return this._compare(Polymer.Base.get(sort.path, a), Polymer.Base.get(sort.path, b));
+          return this._compare(Polymer.Base.get(sort.path, sa), Polymer.Base.get(sort.path, sb));
         } else if (sort.direction === 'desc') {
-          return this._compare(Polymer.Base.get(sort.path, b), Polymer.Base.get(sort.path, a));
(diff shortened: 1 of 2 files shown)


YuriyVaadin:
Dear @nexuslifeline, please consider using sorting with data provider. You can find an example in `Live Demo` of `vaadin-grid`: https://cdn.vaadin.com/vaadin-grid/3.0.2/demo/sorting-filtering.html#sorting-with-data-provider

nexuslifeline:
Hi @YuriyVaadin , thank you for your response. My understanding is, this would be very useful If you have control on your data provider(or any options to change the sort order) as it is giving the responsibility of data sorting to your data provider. Like in my case, my data provider sort data by id and I do not have any sorting options when requesting data to those api. That's why I'm handling the sorting when I receive those data, I actually added just additional details to be exposed on the sort-changed event(which is path and direction) so that I would know how I will going to sort the data. Then, I also notice that this sort-changed event fires 2 times when sort order is changed caused by NULL values not being trapped. Thanks.

tomivirkki:
Thanks for the contribution and apologies for not replying earlier.
Closing this PR for now, notes:
The CI tests for this PR don't seem to be passing.
Other than that, the change seems to cause some unwanted side-effects:
1. It removes the third state from sorters
2. It breaks sorting because items are converted to string literals before comparing their sub-properties (this can be witnessed in the very first sorting demo)

PR #1070, opened 2017-10-30
Fix: Select by itemIdPath

### LINKED ISSUE #899: _isSelected check doesn't work for paginated displays
The `_isSelected` function uses `this.selectedItems.indexOf(item) > -1` which will always be false when the items replaced during page change. This is because new objects are created and this check is checking by reference.

 It would be good to do an actual object comparison here or provide a way to check a certain property for comparison.

stramel:
Fixes #899

**TODO:**

- [x] JSDocs
- [ ] Tests

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1070)

### DIFF
--- a/vaadin-grid-selection-mixin.html
+++ b/vaadin-grid-selection-mixin.html
@@ -16,13 +16,16 @@ This program is available under Apache License Version 2.0, available at https:/
     static get properties() {
       return {
         /**
-        * An array that contains the selected items.
-        */
+         * An array that contains the selected items.
+         */
         selectedItems: {
           type: Object,
           notify: true,
           value: () => []
         },
+        /**
+         * Key to use for object comparison
+         */
         itemIdPath: String,
       };
     }
--- a/vaadin-grid-selection-mixin.html
+++ b/vaadin-grid-selection-mixin.html
@@ -22,7 +22,8 @@ This program is available under Apache License Version 2.0, available at https:/
           type: Object,
           notify: true,
           value: () => []
-        }
+        },
+        itemIdPath: String,
       };
     }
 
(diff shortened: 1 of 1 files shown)


tomivirkki:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-grid-selection-mixin.html, line 29 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1070#-KxlKjFnmO-Ek1Z4uZXy:-KxlKjFnmO-Ek1Z4uZXz:b-anld86) ([raw file](https://github.com/vaadin/vaadin-grid/blob/5acddb275dc4b7cac0d90b8544c41b4654249702/vaadin-grid-selection-mixin.html#L29)):*
> ```HTML
>          * Key to use for object comparison
>          */
>         itemIdPath: String,
> ```

Are you sure you're using the latest master? `itemIdPath` is already defined at https://github.com/vaadin/vaadin-grid/blob/master/vaadin-grid-data-provider-mixin.html#L135

---

*[vaadin-grid-selection-mixin.html, line 43 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1070#-KxlJlPbG-cdFPdgTjfy:-KxlJlPbG-cdFPdgTjfz:b-punecl) ([raw file](https://github.com/vaadin/vaadin-grid/blob/5acddb275dc4b7cac0d90b8544c41b4654249702/vaadin-grid-selection-mixin.html#L43)):*
> ```HTML
>       
>       return precondition && (
>         (this.itemIdPath && this.selectedItems.some((i) => i[this.itemIdPath] === item[this.itemIdPath])) ||
> ```

`item[this.itemIdPath]` doesn't actually look deeper than one level in the object structure. For example. `item['name.first']` would look for a property `name.first` from the item directly instead of looking for `first` under `item.name`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1070)*

stramel:
Ahh, thanks, @tomivirkki I must have missed that.

Issue #1058, opened 2017-10-26
Sorting with empty data cells does not work correctly under version 1.2


bekihm:
Currently the grid contains as data for empty cells the 'value' undefined. Internally vaadin-grid compares two cells a and b with the < and > operator. One problem is that some value > undefined and some value < undefined is always false. This could lead to missorting of the grid.
One possible workaround is to store an empty string as value instead of undefined. So it applies:
some value > ''  => true     and
some value < '' => false

A better solution would be to modify the [compare function used in the method '_sortOrderChanged'](https://github.com/vaadin/vaadin-grid/blob/1.x/vaadin-grid.html#L1246) to something like this:
```
      function _compare(a, b) {
        if(a===undefined && b !== undefined){
          return 1;
        }
        if(b===undefined && a !== undefined){
          return -1;
        }
        if(a===undefined && b===undefined){
          return 0;
        }
        return (a < b ? -1 : a > b ? 1 : 0);
      }
```
Note additionally that Array.sort which is used from vaadin-grid is not stable: 
https://en.wikipedia.org/wiki/Sorting_algorithm#Stability
So vaadin-grid does not keep the other columns sorting while sorting one column multiple times.

web-padawan:
The corresponding [code](https://github.com/vaadin/vaadin-grid/blob/551035642773df4fe23004ab066e13fc20795ad0/src/vaadin-grid-array-data-provider-mixin.html#L127-L138) has been re-implemented to take empty values into account.

We no longer support `vaadin-grid` version 1, and now it is very outdated. Please consider using the most recent version (currently 5.x) which gets all the fixes.

Closing as won't fix.

PR #1121, opened 2017-11-21
1058: Respect null and undefined when sorting columns

newlukai:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1121)


newlukai:
FIxes #1058 .

Issue #1081, opened 2017-11-04
v3 has bad dependencies


JosefJezek:
https://github.com/vaadin/vaadin-grid/blob/v3.0.2/bower.json#L49

Correct dependencies are
```
"iron-a11y-keys-behavior": "1 - 2",
 "polymer": "1.9 - 2",
...
```

vicsstar:
@JosefJezek the dependencies have been updated.

web-padawan:
Fixed in #1059 and released in v3.1.0

PR #1122, opened 2017-11-21
1058: Respect null and undefined when sorting columns

### LINKED ISSUE #1058: Sorting with empty data cells does not work correctly under version 1.2

newlukai:
Fixes #1058

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1122)

### DIFF
--- a/vaadin-grid.html
+++ b/vaadin-grid.html
@@ -1244,6 +1244,19 @@ See the [demo](demo/index.html) for use case examples.
       var evt = this.fire('sort-order-changed', {value: sortOrder}, {cancelable: true});
 
       function _compare(a, b) {
+        var isANull = typeof a === 'undefined' || a === null;
+        var isBNull = typeof b === 'undefined' || b === null;
+
+        if (isANull && !isBNull) {
+          return 1;
+        }
+        if (!isANull && isBNull) {
+          return -1;
+        }
+        if (isANull && isBNull) {
+          return 0;
+        }
+
         return (a < b ? -1 : a > b ? 1 : 0);
       }
 


newlukai:
Seems that the Travis build is fine, but the tests fail due to missing Sauce credentials. How could I fix that?

web-padawan:
@newlukai Currently, our team is inclined not to introduce features to the 1.x branch.
Would it be an option for you to use 3.x instead, which is a hybrid version? 
If not, we might consider this fix, but please add tests for it.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1122#-:-KzTI9WTCqx9ALWPTC8Z:b-hpwnzm)*

newlukai:
@web-padawan: You make a good point, I didn't find any information about which vaadin-grid version is compatible to which Polymer version. Vaadin-grid 2.x already includes the change I proposed here, but is it able to run with Polymer 1.x (we still didn't manage to migrate)?
If yes, I would consider to update.

web-padawan:
@newlukai you're right, this definitely needs to be documented. We will take care of that, but as for now, here is the short description:

- 1.x is the Polymer 1.x version based on extending native elements (like `td`), which is no longer an option in Polymer 2
- 2.x is the Polymer 1.x version which dropped internal table structure in favor of being more flexible 
- 3.x is the hybrid version (both 1.x and 2.x) having mostly the same implementation
- 4.x is the pure Polymer 2.x class-based version, having table structure re-implemented under the hood.

I would recommend trying 2.x and 3.x for using with Polymer 1.x - if your project is using Polymer 1.8+ the hybrid version os the better choice. I will also take a look at #1081 to ensure it can be installed properly.

PR #1130, opened 2017-11-27
Disable keyboard navigation on mobile devices

### LINKED ISSUE #1079: Cell gets a keyboard focus indicator on iOS when clicked
screenshot 2017-11-02 22 05 06

tomivirkki:
Fixes #1079

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1130)

### DIFF
--- a/test/keyboard-navigation.html
+++ b/test/keyboard-navigation.html
@@ -288,8 +288,9 @@
     }
 
     const ios = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
+    const android = /android/i.test(navigator.userAgent);
 
-    !ios && describe('keyboard navigation', () => {
+    !ios && !android && describe('keyboard navigation', () => {
       beforeEach(done => {
         const fixtureElements = fixture('default');
         grid = fixtureElements[0];
--- a/vaadin-grid-keyboard-navigation-mixin.html
+++ b/vaadin-grid-keyboard-navigation-mixin.html
@@ -38,6 +38,11 @@ This program is available under Apache License Version 2.0, available at https:/
     ready() {
       super.ready();
 
+      if (this._ios || this._android) {
+        // Disable keyboard navigation on mobile devices
+        return;
+      }
+
       this.addEventListener('keydown', this._onKeyDown);
       this.addEventListener('focusin', this._onFocusIn);
       this.addEventListener('focusout', this._onFocusOut);
--- a/vaadin-grid.html
+++ b/vaadin-grid.html
@@ -119,7 +119,7 @@ This program is available under Apache License Version 2.0, available at https:/
      * and the “Assigning Data” page in the demos.
      *
      * __Note that expanding the tree grid's item will trigger a call to the `dataProvider`.__
-     * 
+     *
      * __Also, note that when using function data providers, the total number of items
      * needs to be set manually. The total number of items can be returned
      * in the second argument of the data provider callback:__
@@ -244,11 +244,6 @@ This program is available under Apache License Version 2.0, available at https:/
       static get properties() {
         return {
 
-          _safari: {
-            type: Boolean,
-            value: /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
-          },
-
           _this: {
             type: Object,
             value: function() {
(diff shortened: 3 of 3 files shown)


web-padawan:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid.html, line 279 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130#-Kzy2ZC6xFUI19E8TSIf:-Kzy2ZC6xFUI19E8TSIg:b-yp5b8h) ([raw file](https://github.com/vaadin/vaadin-grid/blob/dba9a06e0e16bc6eba5604e82ea77182151538bc/vaadin-grid.html#L279)):*
> ```HTML
>           },
> 
>           _android: {
> ```

Is it necessary to have all this stuff in the root component? Why don't we use a mixin? It could be moved to a separate repo in the future, for consistent usage across different components.

Within a mixin, we could also move the check out of the `properties` block in order to only perform it once and use ES6 block scope, e. g. like this:
```
{
  const IS_ANDROID =  /android/i.test(navigator.userAgent);

  window.Vaadin = window.Vaadin || {};

  Vaadin.UaParserMixin = superClass => class UaParserMixin extends superClass {
    static get properties() {
      return {
        _android: {
          type: Boolean,
          readOnly: true,
          value: IS_ANDROID
        }
      };
    }
  }
}
```

Not a blocker for now.

---

*[test/keyboard-navigation.html, line 291 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130#-Kzy1nnB0tQtpeLXTmiC:-Kzy1nnB0tQtpeLXTmiD:bb4qs0k) ([raw file](https://github.com/vaadin/vaadin-grid/blob/dba9a06e0e16bc6eba5604e82ea77182151538bc/test/keyboard-navigation.html#L291)):*
> ```HTML
> 
>     const ios = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
>     const android = /android/i.test(navigator.userAgent);
> ```

It would be nice these checks in one place, like a mixin as I suggested above. BTW, the `ios` check in the test differs from the check in the `vaadin-grid`.

Not a blocker.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130)*

YuriyVaadin:
I like Sergey's suggestions, however there are not blockers.
:lgtm: 

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130#-:-L-0Tg3p1_L1X6k2Rm7a:b-j3q51b)*

tomivirkki:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-grid.html, line 279 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130#-Kzy2ZC6xFUI19E8TSIf:-L-0_QjBRD0N-zWsu0q3:b-wqdos) ([raw file](https://github.com/vaadin/vaadin-grid/blob/dba9a06e0e16bc6eba5604e82ea77182151538bc/vaadin-grid.html#L279)):*
We've actually been planning to create a common parent for all Vaadin elements (maybe a mixin, maybe an element, perhaps the parent could even replace the current `vaadin-element-skeleton`...) since there is more need to centralize some common logic.

We could move these browser checks to the parent at that point. Wouldn't make any changes about this in the scope of this PR though since we haven't decided anything about the parent's structure yet.

> Within a mixin, we could also move the check out of the properties block in order to only perform it once and use ES6 block scope

This PR removes the `value: function() {...}`s from the properties so the values are determined only once for the prototype.

---

*[test/keyboard-navigation.html, line 291 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130#-Kzy1nnB0tQtpeLXTmiC:-L-0cIA6Od1xQrMAdE4j:b-wtha38) ([raw file](https://github.com/vaadin/vaadin-grid/blob/dba9a06e0e16bc6eba5604e82ea77182151538bc/test/keyboard-navigation.html#L291)):*
<details><summary><i>Previously, web-padawan (Sergey Kulikov) wrote…</i></summary><blockquote>

It would be nice these checks in one place, like a mixin as I suggested above. BTW, the `ios` check in the test differs from the check in the `vaadin-grid`.

Not a blocker.
</blockquote></details>

In case of Grid's tests, helpers.html is the common utility. However, these test definitions get run before HTML imports have completed. We can only use the utilities from helpers.html inside the actual test blocks.

Aligned the `ios` value checks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130)*

web-padawan:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1130#-:-L-0krfhhwZ1J2gh9D9e:bnfp4nl)*

PR #1131, opened 2017-11-28
Passing null as filter breaks grid

### LINKED ISSUE #1128: Passing 'null' as filter breaks grid.
### Description
If `null` is passed as a filter value, vaadin-grid throws a `TypeError: Cannot read property 'toString' of null.`

One use case where this bug may surface is when programmatically setting the filter, e.g. from a Redux store reducer. If the developer neglects to check for `null` in the reducer, this runtime error will surface.

### Expected outcome
No error should be thrown, instead, the filter should be set to the empty string.

### Actual outcome
`TypeError: Cannot read property 'toString' of null.`

### Live Demo
https://jsbin.com/hawafen/edit?html,console,output

You'll likely have to break on vaadin-grid-array-data-provider-behavior.html:127 to see the error.

### Steps to reproduce

2. Add `vaadin-grid-filter` element to the column header template.
3. Set filters value to null.
4. Receive error.

## Proposed Solution

Check for `filter == null` in `_filter` method in vaadin-grid-array-data-provider-behavior.html and substitute with a properly-boxed empty string as needed. Happy to provide PR upon request.

### Browsers Affected

- [x] Chrome
- [x] Firefox

bennypowers:
closes #1128 

Please double check the test logic, thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1131)

### DIFF
--- a/test/filtering.html
+++ b/test/filtering.html
@@ -181,6 +181,13 @@
         flushGrid(grid);
       });
 
+      it('should not throw when receiving null filter', done => {
+        grid._filters[0].value = null;
+        Polymer.Base.async(() => {
+          done();
+        });
+      });
+
       it('should filter automatically', () => {
         expect(getBodyCellContent(grid, 0, 0).innerText).to.equal('foo');
         expect(getBodyCellContent(grid, 1, 0).innerText).to.equal('foo');
--- a/test/filtering.html
+++ b/test/filtering.html
@@ -183,6 +183,7 @@
 
       it('should not throw when receiving null filter', done => {
         grid._filters[0].value = null;
+        flushFilters(grid);
         Polymer.Base.async(() => {
           done();
         });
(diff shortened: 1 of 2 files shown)


test/filtering.html
```diff
@@ -181,6 +181,13 @@
         flushGrid(grid);
       });
 
+      it('should not throw when receiving null filter', done => {
+        grid._filters[0].value = null;
```
limonte: Add `flushFilters(grid);` right after this line to invalidate grid filters. Without `flushFilters(grid);` test is passing without your fix in `vaadin-grid-array-data-provider-mixin.html`

tomivirkki:
Looks like this PR got stuck in limbo, sorry about that :lgtm: Will merge this as https://github.com/vaadin/vaadin-grid/pull/1457

PR #1137, opened 2017-11-29
Submit one data request for empty grid

### LINKED ISSUE #1134: Data provider called repeatedly when column filter hides all rows
### Description
When a column filter is used with the Vaadin Grid, which results in zero rows being displayed (nothing matches the filter), then the data provider is called repeatedly (once for every item in the current page).  This results in extreme performance penalties when the data is being lazy loaded over a network.

### Expected outcome
When a column filter is used that results in zero rows being displayed (nothing matches the filter), then the data provider should be called only once for the active page. The data provider will return an empty array in this case, and then the data provider will not be called again.

### Actual outcome
The data provider is called for the active page X times, where X is the lesser of the page size and total size.  For example, if total grid size is 20, then it will get called 20 times.  If total size is 300 and page size is 50, then it will get called 50 times.

### Steps to reproduce
2. Add `vaadin-grid-filter` element to the column header template.
3. Add a console log into the vaadin-array-data-provider-mixin.html file in the _arrayDataProvider function here: https://github.com/vaadin/vaadin-grid/blob/v4.1.0-alpha5/vaadin-grid-array-data-provider-mixin.html#L53.
4. Assign some items to the grid via JS.
5. Open the page in a web browser.
6. Type into the filter input something that does not match any row in the grid.
7. Observe the large number of console logs when the filter input does not match any item in the grid.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE 11
- [x] iOS Safari
- [x] Android Chrome

tomivirkki:
Fixes #1134

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1137)

### DIFF
--- a/test/data-provider.html
+++ b/test/data-provider.html
@@ -288,6 +288,17 @@
         flushGrid(grid);
       });
 
+      it('should request data only once for empty grid', () => {
+        grid.dataProvider = sinon.spy(infiniteDataProvider);
+        grid.size = 10;
+        grid.dataProvider.reset();
+
+        grid.size = 0;
+        grid.clearCache();
+
+        expect(grid.dataProvider.callCount).to.equal(1);
+      });
+
       it('should warn about missing size', () => {
         const _warn = console.warn;
         const spy = console.warn = sinon.spy();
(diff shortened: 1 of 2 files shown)


limonte:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1137#-:-L-6jfwFf0avYdUggaEX:bnfp4nl)*

PR #1147, opened 2017-12-03
Fix filter memory leak

dumbbillyhardy:
I would really like this for 3.x as well, should I make a PR with these changes cherry-picked over?

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1147)

### DIFF
--- a/src/vaadin-grid-filter-mixin.html
+++ b/src/vaadin-grid-filter-mixin.html
@@ -30,8 +30,12 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     _filterChanged(e) {
-      if (this._filters.indexOf(e.target) === -1) {
+      const index = this._filters.indexOf(e.target);
+      if (index === -1) {
         this._filters.push(e.target);
+      }
+      else if (!e.target.value || e.target.value === "") {
+        this._filters.splice(index, 1);
       }
 
       e.stopPropagation();
--- a/test/filtering.html
+++ b/test/filtering.html
@@ -154,6 +154,12 @@
         grid._filters[0].value = '';
       });
 
+      it('should not keep empty filters', function() {
+        const filter = grid._filters[0];
+        filter.value = '';
+        flushFilters(grid);
+        expect(grid._filters).to.not.include(filter);
+      });
     });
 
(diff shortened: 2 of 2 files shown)


vicsstar:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1147)*

vicsstar:
> dd277d6
This needs a rebase.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1147#-:-L2ec0BbR5QGRq2FKyUc:bmp9dxv)*

vicsstar:
Perhaps a cleaner rebase, one without merge commits.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1147#-:-L2ecS21ghtBaB-G-erI:b-atunvr)*

tomivirkki:
Thanks for the PR @billy-hardy, it kind of solves the problem but it still requires the user to clear the filter value before removing it from the DOM. So the following test isn't satisfied:

```
it('should not keep disconnected filters', function() {
  const filter = grid._filters[0];
  // Filter gets removed from the dom
  filter.parentNode.removeChild(filter);
  flushFilters(grid);
  expect(grid._filters).to.not.include(filter);
});
```

Would it make sense to instead of removing the filter from the array when the value is cleared, to remove it when it gets disconnected.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1147#-:-L3bLgcco9VvZnPKhFGf:b9es3u6)*

tomivirkki:
Closing as inactive, please re-open if this is still valid. I couldn't quite grasp how the current behaviour produces a memory leak or how the proposed change fixes one. The test case checks that a filter with an empty value is removed from the `_filters` array but doesn't explain why referencing one is a problem.

As stated in https://github.com/vaadin/vaadin-grid/pull/1147#issuecomment-360061565 to make sure there are no unwanted filter instances remaining in the array, they should be cleaned up on disconnect, regardless of whether they have a value or not.

PR #1242, opened 2018-03-06
Detect touch device statically

### LINKED ISSUE #1222: Annoying repeated exception from touch support detection
`ColumnReorderingMixin` defines a `_touchDevice` property that is initialized separately each time a `GridElement` instance is created. The detection logic depends on an exception being thrown, which makes things annoying when `Pause on caught exceptions` is enabled in the browser debugger when trying to pinpoint other problems.

If the obtrusive detection logic cannot be avoided, then it should at least only be run once instead of every time a `GridElement` is initialized.

Using version 5.0.0.beta1.

tomivirkki:
Fixes #1222

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1242)

### DIFF
--- a/src/vaadin-grid-column-reordering-mixin.html
+++ b/src/vaadin-grid-column-reordering-mixin.html
@@ -28,17 +28,6 @@ This program is available under Apache License Version 2.0, available at https:/
         _orderBaseScope: {
           type: Number,
           value: 10000000
-        },
-
-        _touchDevice: {
-          value: () => {
-            try {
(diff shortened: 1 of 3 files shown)


web-padawan:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-grid/1242#-:-L6v_8lDzu9h0sIVBKGt:bnfp4nl)*

PR #1245, opened 2018-03-07
Combine mixins to work in Chrome 65

### LINKED ISSUE #1234: "Aw, Snap!" in Chrome Canary 66.0.3352.0
### Description
Chrome Canary 66.0.3352.0 dies with an "Aw, Snap!" page error attempting to load the [vaadin-grid 4.1.5 demos](https://vaadin.com/elements/vaadin-grid/html-examples/grid-basic-demos). I have seen this error with my own use of vaadin-grid beginning with Chrome Canary 66.0.3349.0. The demos (and my app) run fine in Chrome 64.

### Expected outcome
The demos page should load and the demos run.

### Actual outcome
Chrome Canary dies with 
"**Aw, Snap!**
Something went wrong while displaying this webpage."

### Live Demo
Example: https://vaadin.com/elements/vaadin-grid/html-examples/grid-basic-demos

### Steps to reproduce
Just open the demo. The app dies immediatel.

### Browsers Affected

- [ ] Chrome release
- [x] Chrome Canary
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

web-padawan:
Fixes #1234

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1245)

### DIFF
--- a/src/vaadin-grid-column-reordering-mixin.html
+++ b/src/vaadin-grid-column-reordering-mixin.html
@@ -52,7 +52,7 @@ This program is available under Apache License Version 2.0, available at https:/
     ready() {
       super.ready();
       Polymer.Gestures.addListener(this, 'track', this._onTrackEvent);
-      this._reorderGhost = this.$$('[part="reorder-ghost"]');
+      this._reorderGhost = this.shadowRoot.querySelector('[part="reorder-ghost"]');
 
       this.addEventListener('touchstart', this._onTouchStart.bind(this));
       this.addEventListener('touchmove', this._onTouchMove.bind(this));
--- a/src/vaadin-grid-combined-mixin.html
+++ b/src/vaadin-grid-combined-mixin.html
@@ -1,0 +1,828 @@
+<!--
+@license
+Copyright (c) 2017 Vaadin Ltd.
+This program is available under Apache License Version 2.0, available at https://vaadin.com/license/
+-->
+
+<script>
+  window.Vaadin = window.Vaadin || {};
+  window.Vaadin.Grid = window.Vaadin.Grid || {};
+
+  /**
+   * @polymerMixin
+   * TODO: remove this once the Chrome bug gets fixed: https://bugs.chromium.org/p/chromium/issues/detail?id=817499
+   * There is a "limit" of 25 mixins allowed to apply, and Chrome crashes when exceeding this number
+   */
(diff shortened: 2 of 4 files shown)


tomivirkki:
LGTM

moderndeveloperllc:
@web-padawan Just curious - how did you figure this out? The Chromium bug report has a mention of 25+ mixins causing an issue, but `vaadin-grid` only had 16. Was the number you combined arbitrary or anything?

web-padawan:
@moderndeveloperllc actually the `Polymer.Element` brings following mixins:

- PropertiesChanged
- PropertyAccessors
- TemplateStamp
- PropertyEffects
- PropertiesMixin
- ElementMixin

... and since we are also using behaviors, that means the `LegacyElementMixin` usage which turns out to include another bunch of mixins:

- GestureEventListeners
- Dir
- LegacyElement
- MutableData
- OptionalMutableData

My initial idea was to get rid of the `LegacyElementMixin` but that would mean copying the external code (`IronResizableBehavior` etc), and anyways would not be sufficient.

So, actually, I combined the exact number of mixins needed to mitigate the issue + 1 (just in case if someone is brave enough to extend the `vaadin-grid` itself).

moderndeveloperllc:
@web-padawan Thanks for the explanation. I hadn't considered the `Polymer.Element` mixins. FWIW, the `vaadin-grid` 5.0 branch has been working for be very well. It seems to render faster and be more responsive to scrolls for very long lists where I'm not setting the height  - just doing `heightByRows`.

karsten-eger:
> So, actually, I combined the exact number of mixins needed to mitigate the issue + 1 (just in case if someone is brave enough to extend the vaadin-grid itself).

So the consumer of vaadin-grid "only" have one mixin available now? Should be fine for now I guess - at least Chrome is fixing this


already upgraded on our end, it works again, thanks for the quick solution.

web-padawan:
@karsten-eger we tried not to collapse everything into one mixin, and to keep at least some of the biggest untouched. Hope that we will be able to get rid of this workaround before going to stable `5.0.0`, but there still might be some changes until that time, e. g. there is a PR for `column-reordering-mixin` in review as of now, etc.

PR #1328, opened 2018-07-25
Fix tests timeouts in Safari

web-padawan:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1328)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,14 +15,28 @@ before_script:
   - npm install -g bower polymer-cli@next
   - bower install
 
-env:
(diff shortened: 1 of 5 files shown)


web-padawan:
Looks like `animationend` is buggy in Safari 11. Investigating. 
Seems related to 0666c0d which no longer appears to help 😞

web-padawan:
It appears that removing `transform: translateZ(0)` hack (used for some old version of Safari) fixes the bug for Safari 11.1 but need to double check if everything works, waiting for tests first.

PR #1337, opened 2018-07-31
Update demos to prefer renderers

### LINKED ISSUE #1331: Update demos to prefer renderers
Child of vaadin/components-team-tasks#386

YuriyVaadin:
Fixes #1331 

Some of the demos cannot be easily converted and requires `renderers` improvements:
- Exposing `selected` property to `model` (row related)
- Possibility to call `renderer` externally with additional argument/method
- Exposing `level` and `expanded` for `tree-demos` as currently it's working with cell instance

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1337)

### DIFF
--- a/demo/grid-basic-demos.html
+++ b/demo/grid-basic-demos.html
@@ -6,8 +6,66 @@
       }
     </style>
 
+    <h3>Content Renderer Function</h3>
+    <vaadin-demo-snippet id='grid-columns-demos-renderer-functions'>
+      <template preserve-content>
+        <iron-ajax auto url="https://demo.vaadin.com/demo-data/1.0/people?count=200"></iron-ajax>
 
-    <h3>Basic Binding</h3>
+        <vaadin-grid aria-label="Content Renderer Function">
+          <vaadin-grid-column id="column1" width="60px" flex-grow="0"></vaadin-grid-column>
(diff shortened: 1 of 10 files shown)


tomivirkki:
___
*[demo/grid-crud-demos.html, line 153 at r4](https://reviewable.io/reviews/vaadin/vaadin-grid/1337#-LJJ9ySv1kylJcFb0RiE:-LJJ9ySv1kylJcFb0RiF:bkpb2x0) ([raw file](https://github.com/vaadin/vaadin-grid/blob/3cdb691d3f6597f563060d7c95e68c69a60076b4/demo/grid-crud-demos.html#L153)):*
> ```HTML
>             function getTextField(column, index) {
>               const cell = column._allCells.filter(function(cell) {
>                 return cell._model && cell._model.index === index;
> ```

Private API use in the demos, not ok

tomivirkki:
I'd prefer the approach that was used previously in the demo (marking the elements with an identifier and fetching them by that id once needed) instead of adding new API https://github.com/vaadin/vaadin-grid/issues/1352

The proposed API isn't consistent either because it returns different results depending on the user's scroll position/number of rows iron-list renders to the dom etc. For example, if you're scrolled to row 1000, what should `grid.getCellContent(0, 0)` return?

PR #1400, opened 2018-09-10
Prefer column convenience properties in the demos [skip ci]

### LINKED ISSUE #492: Convenience <vaadin-grid-column> properties
Add the following properties to `<vaadin-grid-simple-column>` for convenient configuration:
- `path="name.first"` 
  - No need to provide a body-template, grid should use a default one that renders the path value as a String)
  - No need to provide a header-template, grid generates a default header label (in this case "First")
- `label="First Name"`
  - No need to provide a header-template, the label String is used
- `sortable`
  - No need to provide a vaadin-grid-sorter, the value is used as the path for a default sorter (if empty, use the `path` attribute/property if defined)
- `filterable`
  - No need to provide a vaadin-grid-filter, the value is used as the path for a default filter

### LINKED ISSUE #1384: Update demos to prefer column convenience properties
Also update the README example and API docs examples

Prototype https://github.com/vaadin/vaadin-grid/tree/simple-column

child of #492
needs #1388

tomivirkki:
Fixes #1384
Closes #492

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1400)


jouni:
Nice, a massive improvement! Congrazt for the great work 🎉

PR #1444, opened 2018-09-27
Add `first-column` attribute on the first visible column

jouni:
Enables theming to target the first column in the grid.

I think this is a small enhancement to the current “theming” feature, and could be released in a patch release. Not sure how strict our semver rules are, if this is considered a new feature and would require a minor version.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1444)

### DIFF
--- a/src/vaadin-grid-column-reordering-mixin.html
+++ b/src/vaadin-grid-column-reordering-mixin.html
@@ -298,7 +298,7 @@ This program is available under Apache License Version 2.0, available at https:/
       column1._order = column2._order;
       column2._order = _order;
       this._updateLastFrozen();
-      this._updateLastColumn();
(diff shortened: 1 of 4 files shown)


jouni:
So, is this a patch or minor version? Does grid already use semantic-release or not – how should this PR be tagged?

web-padawan:
@jouni we so not use semantic-release for components at the moment.
It should be ok to include this feature in the current alpha period, WDYT?

tomivirkki:
We can include this in the ongoing minor/alpha period IMO

limonte:
@jouni released as [v5.2.0-alpha5](https://github.com/vaadin/vaadin-grid/releases/tag/v5.2.0-alpha5) 🎉

PR #1446, opened 2018-09-28
Truncate inline content with ellipsis

jouni:
The upcoming [`<vaadin-crud>` component makes this change in its own styles](https://github.com/vaadin/vaadin-crud/blob/843f87b8316db2e65e28add15d37e4e99c151d9f/theme/lumo/vaadin-crud-styles.html#L28-L30), but I believe it should be part of the core grid styles. Previously I didn’t add this because it caused Chrome to show the ellipsis also for the selection column checkboxes, and I didn’t take the time to debug the root cause(s).

Increase the default width of the selection column so that it will not get an ellipsis in Chrome.

Remove empty text nodes from the selection column checkboxes, as that will increase the width of the checkbox and again cause an ellipsis to show (again in Chrome).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1446)


tomivirkki:
*[a discussion](https://reviewable.io/reviews/vaadin/vaadin-grid/1446#-LNjs15xCB_k9UtEpQCK:-LNjs15xCB_k9UtEpQCL:b-fs9prp) (no related file):*
Doesn't seem to pass "Expected empty line before rule"

---

jouni:
That lint error was actually introduced in my [previous commit](https://github.com/vaadin/vaadin-grid/commit/b638a5e68549e98d6ba8539f217850420e0e6a45) which I merged directly to master (as it only touched on the docs).

tomivirkki:
*[a discussion](https://reviewable.io/reviews/vaadin/vaadin-grid/1446#-LO28JgW8NtAC6DaaBgC:-LO28JgXELJB6Ue181zM:b8kfbgu) (no related file):*
Some of the demos could use an update to the column widths https://www.dropbox.com/s/36rpc101n81mqrq/Screenshot%202018-10-05%2010.33.08.png?dl=0

Also the very first demo has: "Visit Co..." as a column header

---

src/vaadin-grid-selection-column.html
```diff
@@ -12,12 +12,10 @@
   <template>
     <template class="header" id="defaultHeaderTemplate">
       <vaadin-checkbox class="vaadin-grid-select-all-checkbox" aria-label="Select All" hidden$="[[_selectAllHidden]]" on-checked-changed="_onSelectAllCheckedChanged"
-        checked="[[_isChecked(selectAll, _indeterminate)]]" indeterminate="[[_indeterminate]]">
-      </vaadin-checkbox>
+        checked="[[_isChecked(selectAll, _indeterminate)]]" indeterminate="[[_indeterminate]]"></vaadin-checkbox>
```
limonte: `vaadin-checkbox` `2.2.3` has been released, no need in this change anymore.

src/vaadin-grid-selection-column.html
```diff
@@ -12,12 +12,10 @@
...
     </template>
     <template id="defaultBodyTemplate">
-      <vaadin-checkbox aria-label="Select Row" checked="{{selected}}">
-      </vaadin-checkbox>
+      <vaadin-checkbox aria-label="Select Row" checked="{{selected}}"></vaadin-checkbox>
```
limonte: `vaadin-checkbox` `2.2.3` has been released, no need in this change anymore.

src/vaadin-grid-selection-column.html
tomivirkki: Grid still depends on `^2.2.1` so in theory this should work with version `2.2.1` also. Imo the change is fine.

src/vaadin-grid-selection-column.html
tomivirkki: Grid still depends on `^2.2.1` so in theory this should work with version `2.2.1` also. Imo the change is fine.

tomivirkki:
*[a discussion](https://reviewable.io/reviews/vaadin/vaadin-grid/1446#-LO28JgW8NtAC6DaaBgC:-LQDy2auEgHei7jsHkRB:b-896fix) (no related file):*

Some of the demos could use an update to the column widths https://www.dropbox.com/s/36rpc101n81mqrq/Screenshot%202018-10-05%2010.33.08.png?dl=0

Also the very first demo has: "Visit Co..." as a column header
</blockquote></details>

Done.

---

Issue #1468, opened 2018-10-15
Calling `notifyPath` or `notifySplices` on element or parent tempalte causes object to be changed again.


tomalec:
### Description
I use Polymer's `notifyPath` and `notifySplice` methods on `dom-bind`/parent element to notify Polymer notification protocol about the changes made to the view-model by external system.

Also I use JS `Proxy` to observe my data object and propagate the changes further.

`vaadin-grid` with header `template` performs the change again instead of notifying the binding. That is extremely undesired when it comes to splice operations.

### Expected outcome
When `notifyPath` is called on parent polymer element, header template should be notified, data object should remain untouched.

### Actual outcome
Header template changes original object once again.

### Live Demo
https://glitch.com/edit/#!/material-louse?path=index.html:25:17

### Steps to reproduce
1. Put a `vaadin-grid` element in any Polymer element, like `dom-bind`,
2. Add `template class="header` with any data binding,
3. Call `parentElement.notifyPath('some.path', 'somevalue')`;

### Browsers Affected
- [x] Chrome
- [ ] Firefox
- [?] Safari
- [ ] Edge
- [?] IE 11
- [?] iOS Safari
- [?] Android Chrome

PR #1469, opened 2018-10-15
Notify templateInstances instead of setting the value again,

tomalec:
fixes https://github.com/vaadin/vaadin-grid/issues/1468

I was not sure how you structure tests. Do you want me to add test case in `/test/templates.html` or somewhere else?

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1469)

### DIFF
--- a/src/vaadin-grid-templatizer.html
+++ b/src/vaadin-grid-templatizer.html
@@ -140,7 +140,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
       _forwardParentProp(prop, value) {
         this._parentPathValues[prop] = value;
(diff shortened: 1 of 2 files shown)


tomivirkki:
Hi, thanks for the PR. A test to validate it would be great, `/test/templates.html` is a good place for it 👍

tomalec:
Test added.

Sorry for verbose spy assignment. I was not able to use Sinon's way to spy on setters `spy = sinon.spy(model, 'foo', ['set'])`. Maybe that's due to outdated Sinon in WCT.

Issue #1256, opened 2018-03-13
modelForElement


robrez:
### Description
Would like a `modelForElement` function or similar

### Expected outcome
```javascript
_onSomeEvent(e) {
  let model = this.$.grid.modelForElement(e.target);
}
```

### Actual outcome
```javascript
_onSomeEvent(e) {
  const path = e.composedPath();
  const cell = path.filter(item => item._instance && item.getAttribute("role") === "gridcell");
  if (cell) {
    let model = cell._instance;
  }
}
```

moderndeveloperllc:
@robrez This might not be what you are looking for, but you can get `e.model.item` for events. It pulls in the item for that iteration. I currently do something similar to
```
_passwordReset(event) {
    let resetButton = event.path[2];
    let username = event.model.item.username;
    // Logic here
}
```

robrez:
@moderndeveloperllc Thanks; IIRC, that does work when you add event handlers to individual templates instances:

```html
<v-grid>
  <v-grid-column>
    <template class="header">Name</template>
    <template><div on-blah="_onBlah">[[item.name]]</div></template>
  </v-grid-column>
</v-grid>
```

I was trying to have a single event listener:

```html
<v-grid on-blah="_onBlah">
  <v-grid-column>
    <template class="header">Name</template>
    <template>[[item.name]]</template>
  </v-grid-column>
</v-grid>
```

tomivirkki:
Thanks for the suggestion @robrez Something like this would be handy.

One issue needing to be resolved is that since the default theme adds paddings to the `td` elements, there's currently no way to listen to mouse events that occur over the area covered by the paddings. Even if you add click listeners to your content elements, clicks over the padding are are ignored.

This manifests in the grid demo at https://cdn.vaadin.com/vaadin-context-menu/4.1.0-alpha1/demo/#context-menu-advanced-demos


Thinking out loud:
```
grid.addEventListener('click', e => {
 const model = grid.modelForEvent(e);
)
```
...dunno, looks hairy. In this case the click could come from anywhere inside the grid.

Perhaps an additional custom event?
```
grid.addEventListener('body-cell-click', e => {
 const model = e.detail;
)
```

robrez:
That does look convenient. 

Fwiw, we also attach dblclick listeners to some grids. You may not want to add all of these variations of internal listeners?

pekam:
The `detailedEvents` feature #1473 covers these use-cases, so I'm closing this issue.
[...]

PR #1473, opened 2018-10-23
Add API for detailed events

### LINKED ISSUE #1318: Add API for getting the target item from a mouse event
To make it easier to use `<vaadin-context-menu>` with `<vaadin-grid>`, it would be nice to have public API for the logic of retrieving the item based on the clicked row, that is used for selection:
https://github.com/vaadin/vaadin-grid/blob/6d78144cb760c80f914e5dd4312c42c01a67444c/src/vaadin-grid-cell-click-mixin.html#L24

Otherwise I need to copy this logic for the Flow component implementation:
```javascript
grid.addEventListener('vaadin-contextmenu', function(e) {
      const path = e.composedPath();
      const cell = path[path.indexOf(grid.$.table) - 3]; // <td> element in shadow dom
      var key;
      if (cell && cell._instance.item) {
        key = cell._instance.item.key;
      }
      grid.$server.updateContextMenuTargetItem(key);
    });
```
Now if the web component changes, it breaks the Flow component until we make the same changes for our code.

It could also improve the frontend DX for using `<vaadin-context-menu>` with `<vaadin-grid>`. The current demo adds the item index to the cell contents, which is not very convenient: https://cdn.vaadin.com/vaadin-context-menu/4.1.0-alpha1/demo/#context-menu-advanced-demos

pekam:
Fix #1318, maybe also #1256

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1473)

### DIFF
--- a/demo/demos.json
+++ b/demo/demos.json
@@ -101,6 +101,17 @@
   }
   ,
   {
+    "name": "Detailed Events",
+    "url": "grid-detailed-event-demos",
+    "src": "grid-detailed-event-demos.html",
+    "meta": {
+      "title": "vaadin-grid – Detailed Events",
+      "description": "",
+      "image": ""
+    }
+  }
+  ,
+  {
     "name": "Tree Grid",
     "url": "grid-tree-demos",
     "src": "grid-tree-demos.html",
--- a/demo/grid-detailed-event-demos.html
+++ b/demo/grid-detailed-event-demos.html
@@ -1,0 +1,70 @@
+<dom-module id="grid-detailed-event-demos">
+  <template>
+    <style include="vaadin-component-demo-shared-styles">
+      :host {
+        display: block;
+      }
+    </style>
+
+    <style>
+      vaadin-grid {
+        text-transform: none;
+      }
+    </style>
+
+    <h3>Detailed Events</h3>
+    <p>
+      Set <code>detailedEvents</code> array to fire events with useful information about the event target,
+      like the targeted item, column and section (body, header, footer or row details).
+    </p>
+    <p>
+      This feature can be useful to execute item-specific actions, like selection or opening row details
+      on any mouse or key event.
+    </p>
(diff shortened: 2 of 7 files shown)


tomivirkki:
*[a discussion](https://reviewable.io/reviews/vaadin/vaadin-grid/1473#-LPVg7AEA_PJ-9aXjngQ:-LPVwCPzETXj-M0Ocxjo:bwny2gw) (no related file):*

Might be confusing to test an arbitrary list of events. The current tests are enough imo
</blockquote></details>

Anyway, this is the problematic code `

---

tomivirkki:
*[a discussion](https://reviewable.io/reviews/vaadin/vaadin-grid/1473#-LPVg7AEA_PJ-9aXjngQ:-LPVwNAzEj3yXCMCrtH6:bfabngz) (no related file):*

Anyway, this is the problematic code `
</blockquote></details>

...`https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-keyboard-navigation-mixin.html#L73` I believe it's no longer needed and could be removed

---

tomivirkki:
___
*[demo/grid-detailed-event-demos.html, line 41 at r2](https://reviewable.io/reviews/vaadin/vaadin-grid/1473#-LPWDZK3DFQkwHlSw_Fg:-LPWDZK42Rora_6F-94T:b35gq4w) ([raw file](https://github.com/vaadin/vaadin-grid/blob/e0981f4d865c830a611af156910232358bc598ba/demo/grid-detailed-event-demos.html#L41)):*
> ```HTML
>             grid.detailedEvents = ['dblclick', 'keypress'];
> 
>             grid.addEventListener('detailed-dblclick', e => {
> ```

As the demo sources are loaded dynamically, we're required to stick to ES5 (in the demo sources only) in order for the demo to work on IE11

PR #1474, opened 2018-10-24
Don't try to reorder children if pool is not increased

tomivirkki:
This fixes some issues that occurred in the Flow counterpart tests. Note: Couldn't find a proper isolated test case for the fix at "vaadin-grid-a11y-mixin.html".

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1474)


web-padawan:
Merging without Firefox tests passed due to Polymer/tools#751

PR #1488, opened 2018-11-12
Add cellClassNameGenerator API

### LINKED ISSUE #1481: Implement the cell class generator for grid
Child of #1297 

Early prototype at https://github.com/vaadin/vaadin-grid/tree/class-generators2

pekam:
Allows defining a function that adds CSS class names to the cells in grid's body, based on the corresponding `rowData` and column.

Resolves #1481

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1488)

### DIFF
--- a/demo/grid-styling-demos.html
+++ b/demo/grid-styling-demos.html
@@ -32,8 +32,6 @@
               let classes = rowData.item.gender;
               if (column.path === 'email') {
                 classes += ' email';
-              } else if (column.path === 'name.first' && rowData.item.name.title === 'mrs') {
-                classes += ' mrs';
               }
               return classes;
             };
@@ -46,19 +44,15 @@
               <style>
                 /* Background needs a stronger selector to not be overridden */
                 [part~="cell"].female {
-                  background: var(--lumo-primary-color-10pct);
+                  background: rgb(255, 240, 240);
                 }
 
                 [part~="cell"].male {
-                  background: var(--lumo-contrast-5pct);
+                  background: rgb(245, 245, 255);
                 }
 
                 .email {
                   font-style: italic;
-                }
-
-                .mrs {
-                  font-weight: bold;
                 }
               </style>
           </template>
@@ -71,9 +65,8 @@
     <vaadin-demo-snippet id='grid-styling-demos-regenerate-classes'>
       <template preserve-content>
 
-        <vaadin-button style="margin-bottom: 20px" id="change-title-btn">Change Title</vaadin-button>
-
-        <vaadin-grid aria-label="Styling Grid With Cell Class Generator Example" id="grid" size="200">
+        <vaadin-grid aria-label="Updating Cell Class Generator Example" id="grid" size="200">
+          <vaadin-grid-column id="check" flex-grow="0" width="60px"></vaadin-grid-column>
           <vaadin-grid-column path="name.first" header="First name"></vaadin-grid-column>
           <vaadin-grid-column path="name.last" header="Last name"></vaadin-grid-column>
           <vaadin-grid-column path="email"></vaadin-grid-column>
@@ -84,26 +77,57 @@
             const grid = document.querySelector('vaadin-grid');
             grid.items = Vaadin.GridDemo.users;
 
-            const changeTitleBtn = document.querySelector('#change-title-btn');
-            changeTitleBtn.addEventListener('click', function() {
-              grid.items[0].name.title = grid.items[0].name.title === 'ms' ? 'mrs' : 'ms';
+            const checkColumn = document.querySelector('#check');
+            checkColumn.renderer = function(root, column, rowData) {
+              if (!root.firstElementChild) {
+                root.innerHTML = '<vaadin-checkbox></vaadin-checkbox>';
+                root.firstElementChild.addEventListener('checked-changed', function(e) {
+                  root.item.checked = e.detail.value;
 
-              // Classes are re-generated also by clearCache() and render(),
-              // but this function is lighter as it doesn't do anything extra.
-              grid.generateClassNames();
-            });
+                  // Classes are re-generated also by clearCache() and render(),
+                  // but this function is lighter as it doesn't do anything extra.
+                  grid.generateClassNames();
+                });
+              }
+              root.item = rowData.item;
+              root.firstElementChild.checked = rowData.item.checked;
+            };
 
             grid.cellClassNameGenerator = function(rowData, column) {
               let classes = rowData.item.gender;
               if (column.path === 'email') {
                 classes += ' email';
-              } else if (column.path === 'name.first' && rowData.item.name.title === 'mrs') {
-                classes += ' mrs';
+              } else if (column.path === 'name.first' && rowData.item.checked) {
+                classes += ' checked';
               }
               return classes;
             };
           });
         </script>
+
+        <!-- Place in index.html or in a separate HTML import -->
+        <dom-module id="my-grid-styles" theme-for="vaadin-grid">
+          <template>
+              <style>
+                /* Background needs a stronger selector to not be overridden */
+                [part~="cell"].female {
+                  background: rgb(255, 240, 240);
+                }
+
+                [part~="cell"].male {
+                  background: rgb(245, 245, 255);
+                }
+
+                .email {
+                  font-style: italic;
+                }
+
+                .checked {
+                  font-weight: bold;
+                }
+              </style>
+          </template>
+        </dom-module>
       </template>
     </vaadin-demo-snippet>
 
--- a/demo/grid-styling-demos.html
+++ b/demo/grid-styling-demos.html
@@ -46,7 +46,7 @@
               <style>
                 /* Background needs a stronger selector to not be overridden */
                 [part~="cell"].female {
-                    background: var(--lumo-primary-color-10pct);
+                  background: var(--lumo-primary-color-10pct);
                 }
 
                 [part~="cell"].male {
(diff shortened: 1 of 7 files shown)


test/class-name-generator.html
```diff
@@ -0,0 +1,116 @@
...
+
+<body>
+
+  <script>
+    describe('class name generator', () => {
+      let grid, firstCell, initialCellClasses;
+
+      beforeEach(() => {
+
+        grid = document.createElement('vaadin-grid');
```
tomivirkki: Would still use text-fixture for now. The whole "no-fixture initiative" is in an early stage and the current take is that we could just replace the import from all of our tests to use Vaadin's test fixture.

demo/index.html
```diff
@@ -24,6 +24,30 @@
   <link rel="import" href="../../iron-ajax/iron-ajax.html">
   <link rel="import" href="../../vaadin-lumo-styles/icons.html">
 
+  <!-- Used in the styling examples, keep in sync -->
+  <dom-module id="grid-styling-demo-theme" theme-for="vaadin-grid">
+    <template>
+        <style>
+          /* Background needs a stronger selector to not be overridden */
+          [part~="cell"].female {
+              background: var(--lumo-primary-color-10pct);
```
tomivirkki: Wrong indent

src/vaadin-grid-styling-mixin.html
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Runs the `cellClassNameGenerator` for the visible cells.
+     * If the generator depends on varying conditions, you need to
+     * call this function manually in order to update the styles when
+     * the conditions change.
+     */
+    generateClassNames() {
+      Array.from(this.$.items.children).forEach(
```
tomivirkki: Note that iron-list items container may include hidden rows, for example after the items size is reduced from 2 to 1. So `this.$.items.children` should be filtered before processing.

demo/index.html
```diff
@@ -24,6 +24,30 @@
...
+          }
+
+          [part~="cell"].male {
+            background: var(--lumo-contrast-5pct);
```
tomivirkki: Let's avoid using Lumo properties in non-Lumo-specific demos

demo/grid-styling-demos.html
```diff
@@ -0,0 +1,119 @@
...
+        </dom-module>
+      </template>
+    </vaadin-demo-snippet>
+
+    <h3>Updating Styles After Changes in the Data</h3>
+
+    <vaadin-demo-snippet id='grid-styling-demos-regenerate-classes'>
+      <template preserve-content>
+
+        <vaadin-button style="margin-bottom: 20px" id="change-title-btn">Change Title</vaadin-button>
```
tomivirkki: Just looking at the demo, it's a bit unclear that clicking a global "Change Title" button changes the marital status of the first item in the grid.

src/vaadin-grid-styling-mixin.html
```diff
@@ -0,0 +1,73 @@
...
+        row => this._generateClassNames(row, this.__getRowModel(row)));
+    }
+
+    _generateClassNames(row, rowData) {
+      Array.from(row.children).forEach(cell => {
+        if (cell.__generatedClasses) {
```
tomivirkki: Inspecting the DOM, looks like the `td` -elements have no default class names at all. Could this loop be simplified to just
```
cell.className = '';
this.cellClassNameGenerator && (cell.className = this.cellClassNameGenerator(rowData, cell._column));
```

test/class-name-generator.html
pekam: Done

demo/index.html
pekam: Done

src/vaadin-grid-styling-mixin.html
pekam: Done

src/vaadin-grid-styling-mixin.html
pekam: On polyfilled browsers some classes are added to the `td`-elements, so we can't just clear them.

tomivirkki: Right, forgot about the shady css classes.

demo/index.html
pekam: Done

demo/grid-styling-demos.html
pekam: Fixed: Now the demo has a column of checkboxes to change the styles on any row.

src/vaadin-grid-styling-mixin.html
```diff
@@ -0,0 +1,73 @@
...
+         * - `rowData` The object with the properties related with
+         *   the rendered item, contains:
+         *   - `rowData.index` The index of the item.
+         *   - `rowData.item` The item.
+         *   - `rowData.expanded` Sublevel toggle state.
+         *   - `rowData.level` Level of the tree represented with a horizontal offset of the toggle button.
+         *   - `rowData.selected` Selected state.
+         * - `column` The `<vaadin-grid-column>` element.
+         */
+        cellClassNameGenerator: Function
```
tomivirkki: These new APIs aren't shown in the API doc viewer (analysis.json generates incorrectly)

pekam: Fixed (also the same mistake I've made for `detailedEvents`)

demo/grid-styling-demos.html
```diff
@@ -0,0 +1,143 @@
...
+          window.addDemoReadyListener('#grid-styling-demos-regenerate-classes', function(document) {
+            const grid = document.querySelector('vaadin-grid');
+            grid.items = Vaadin.GridDemo.users;
+
+            const checkColumn = document.querySelector('#check');
+            checkColumn.renderer = function(root, column, rowData) {
+              if (!root.firstElementChild) {
+                root.innerHTML = '<vaadin-checkbox></vaadin-checkbox>';
+                root.firstElementChild.addEventListener('checked-changed', function(e) {
+                  root.item.checked = e.detail.value;
```
tomivirkki: Rather than changing the provided data (not necessarily ok), we could have a `const checkedItems = []` array or an object in the demo code which would host the state of the checked items. Currently, as the data comes from a shared source, any changes to it are effective in the other demos as well.

pekam: Done

demo/index.html
```diff
@@ -24,6 +24,30 @@
   <link rel="import" href="../../iron-ajax/iron-ajax.html">
   <link rel="import" href="../../vaadin-lumo-styles/icons.html">
 
+  <!-- Used in the styling examples, keep in sync -->
+  <dom-module id="grid-styling-demo-theme" theme-for="vaadin-grid">
+    <template>
+        <style>
```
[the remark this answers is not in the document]
pekam: Done (I wonder why linter doesn't catch these)

src/vaadin-grid-styling-mixin.html
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    static get observers() {
+      return [
+        '__cellClassNameGeneratorChanged(cellClassNameGenerator)'
+      ];
+    }
+
+    __cellClassNameGeneratorChanged(cellClassGenerator) {
+      this.generateClassNames();
```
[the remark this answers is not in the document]
pekam: I would prefer not to do this, because currently you can clear the generated styles by setting `cellClassNameGenerator` to a falsy value (I think there might be use-cases for this).

After the proposed change, the following code would leave the 'foo' classes to the cells:
```javascript
grid.cellClassNameGenerator = (rowData, column) => 'foo';
grid.cellClassNameGenerator = undefined;
```

But maybe I should skip the generating if both the old value and the new value are falsy? In this case it doesn't make any difference for the behavior.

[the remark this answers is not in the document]
web-padawan: You can use single property observer, e.g. instead of `static get observers()` make it looks like this:
```js
cellClassNameGenerator: {
  type: Function,
  observer: '__cellClassNameGeneratorChanged'
}
```
there you could check for both new and old value, to make sure whether property has been reset or not.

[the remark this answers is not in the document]
pekam: Done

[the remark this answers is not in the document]
web-padawan: Note the syntax in my example above, `observer` should be reference to the class method, not the function itself (how it is in your code now). See this example: https://github.com/vaadin/vaadin-grid/blob/bc060564eb2cb09d78ea538dc8436678e4f74930/src/vaadin-grid-data-provider-mixin.html#L86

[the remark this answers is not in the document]
pekam: Can you elaborate why it needs to be done in that way?

I noted the syntax but IMO it's cleaner like this, since the function is so simple and it doesn't need to add a new private function to grid (it's already painful to browse through the grid API).

I got this from the Polymer docs: https://www.polymer-project.org/2.0/docs/devguide/observers#change-callbacks (second example)
so I would expect it to be okay unless we have some common convention about this.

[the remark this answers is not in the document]
pekam: > first time the complex observers are invoked with the undefined passed as an argument.

Actually I tested this, and the observer is not run with `undefined` (or anything else) initially. If I don't set the property, the observer is not called at all. If I set the property, the property value is passed as the argument on the first and only time.

So now my plan is to just revert that commit. Is it okay, or did I maybe misunderstand your original idea in this thread?

Also that observer syntax that I used seems to cause problems after Polymer3-conversion so better not to use it.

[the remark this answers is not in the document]
pekam: I reverted the commit for now.

PR #1489, opened 2018-11-13
Add max-width for sorter

### LINKED ISSUE #1438: Nested grid-sorter + grid-filter overflows a narrow column
Nesting `<vaadin-grid-filter>` inside `<vaadin-grid-sorter>` does not scale the contained filter text field to fit the column width.

Test case: https://github.com/vaadin/vaadin-grid/blob/master/demo/dev.html#L116-L118

The `<vaadin-grid-sorter>` element should have similar `max-width: 100%` defined as `<vaadin-grid-filter>` has.

jouni:
Fixes #1438

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1489)

### DIFF
--- a/src/vaadin-grid-sorter.html
+++ b/src/vaadin-grid-sorter.html
@@ -28,9 +28,14 @@ This program is available under Apache License Version 2.0, available at https:/
         max-width: 100%;
       }
 
+      [part="content"] {
+        flex: 1 1 auto;
+      }
+
       [part="indicators"] {
         position: relative;
         align-self: center;
+        flex: none;
       }
 
       [part="order"] {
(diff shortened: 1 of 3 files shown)


src/vaadin-grid-sorter.html
```diff
@@ -25,6 +25,7 @@
       :host {
         display: inline-flex;
         cursor: pointer;
+        max-width: 100%;
```
tomivirkki: Couldn't get it working nice with just this fix. `[part="content"]` also needs 100% (max) width?

jouni: You’re right, thanks! I’ll fix it.

jouni: That hopefully does it.

PR #1500, opened 2018-11-20
Replace detailedEvents with getEventContext API

### LINKED ISSUE #1497: Replace detailedEvents with getEventContext API
After DX tests and discussions, we have concluded to:
- Remove the `detailedEvents` API
- Add function `getEventContext(event)`, which returns the details object for a native event
- Flatten the properties of `rowData` into the returned object

So instead of this:
```javascript
grid.detailedEvents = ['click'];
grid.addEventListener('detailed-click', e => {
  const item = e.detail.rowData.item;
});
```
The user would write this:
```javascript
grid.addEventListener('click', e => {
  const item = grid.getEventContext(e).item;
})
```

pekam:
Simplify the API for getting information about row, column etc. from a
mouse or keyboard event.

Fix #1497

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1500)


src/vaadin-grid-event-context-mixin.html
```diff
@@ -0,0 +1,60 @@
...
+     * This means mostly mouse and keyboard events. If such a grid part is not found in the path, an empty object is returned.
+     * This may be the case eg. if the event is fired on the `<vaadin-grid>` element and not any deeper in the DOM, or if
+     * the event targets the empty part of the grid body.
+     */
+    getEventContext(event) {
+
+      const context = {};
+
+      const path = event.composedPath();
+      const cell = path[path.indexOf(this.$.table) - 3];
```
web-padawan: this part of logic is already present in `ActiveItemMixin`, could we extract it into internal method, e.g. `_getCellByEvent(event)`?

demo/grid-event-context-demos.html
```diff
@@ -0,0 +1,65 @@
...
+  <template>
+    <style include="vaadin-component-demo-shared-styles">
+      :host {
+        display: block;
+      }
+    </style>
+
+    <style>
+      vaadin-grid {
+        text-transform: none;
```
web-padawan: The whole style block with this `text-transform` rule seems to be an old copy-paste leftover and can be removed.

pekam: Done

src/vaadin-grid-event-context-mixin.html
tomivirkki: In that case `vaadin-grid.html` would be the only common place. If there isn't that much shared code, would prefer not to bloat the main class.

web-padawan: Sounds logical. We can keep at as it is for now.

src/vaadin-grid-event-context-mixin.html
```diff
@@ -0,0 +1,60 @@
...
+     * - `item`: the data object corresponding to the targeted row (not specified when targeting header or footer)
+     * - `column`: the column element corresponding to the targeted cell (not specified when targeting row details)
+     * - `section`: whether the event targeted the body, header, footer or details of the grid
+     *
+     * These additional properties are included when `item` is specified:
+     * - `index`: the index of the item
+     * - `selected`: the selected state of the item
+     * - `detailsOpened`: whether the row details are open for the item
+     * - `expanded`: the expanded state of the tree toggle
+     * - `level`: the tree hierarchy level represented with a horizontal offset of the toggle button
```
tomivirkki: Should not mention "toggle button". This only applies in the `vaadin-tree-toggle`

pekam: I copied that from the `renderer()` docs. Done.

tomivirkki: Ouch, shouldn't be there either

PR #1502, opened 2018-11-22
Unpin webcomponentsjs to test against 1.3.0 and 2.2.0

web-padawan:
The polyfill has been pinned for reasons, PR is to verify whether tests pass in 1.3.0 and 2.2.0 or not.


limonte:
As we verified that the issue wasn't fixed in 2.2.0, should this PR be closed?

PR #1515, opened 2018-12-08
Fix heightByRows resizing

### LINKED ISSUE #1510: Height by rows doesn't work in context menu demos
### Description
Setting `grid.heightByRows = true` for the grid in https://cdn.vaadin.com/vaadin-context-menu/4.2.0/demo/#context-menu-grid-demos expands the grid height but not all the rows get rendered. Could be an issue in the demo.

### Expected outcome
All the 100 items get rendered

### Actual outcome
Only the first ~25 items are rendered while the grid height is expanded beyond this

### Live Demo
https://cdn.vaadin.com/vaadin-context-menu/4.2.0/demo/#context-menu-grid-demos

### Steps to reproduce
1. Open https://cdn.vaadin.com/vaadin-context-menu/4.2.0/demo/#context-menu-grid-demos
2. Set `grid.heightByRows = true` with the help of developer tools

### Browsers Affected
Chrome at least

samiheikki:
Fixes #1510

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1515)

### DIFF
--- a/src/vaadin-grid.html
+++ b/src/vaadin-grid.html
@@ -389,7 +389,8 @@ This program is available under Apache License Version 2.0, available at https:/
           heightByRows: {
             type: Boolean,
             value: false,
-            reflectToAttribute: true
+            reflectToAttribute: true,
+            observer: '_heightByRowsChanged'
           }
         };
       }
@@ -713,6 +714,12 @@ This program is available under Apache License Version 2.0, available at https:/
        */
       notifyResize() {
         super.notifyResize();
+      }
+
+      _heightByRowsChanged(value, oldValue) {
+        if (value || oldValue) {
+          this.notifyResize();
+        }
       }
     }
 
--- a/test/resizing.html
+++ b/test/resizing.html
@@ -58,6 +58,16 @@
     </template>
   </test-fixture>
 
+  <test-fixture id="height-by-rows">
+    <template>
+      <vaadin-grid>
+        <vaadin-grid-column>
+          <template>[[index]]</template>
+        </vaadin-grid-column>
+      </vaadin-grid>
+    </template>
+  </test-fixture>
+
   <script>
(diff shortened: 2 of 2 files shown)


src/vaadin-grid.html
```diff
@@ -714,6 +715,10 @@
       notifyResize() {
         super.notifyResize();
       }
+
+      _heightByRowsChanged() {
+        this.notifyResize();
```
web-padawan: Should we guard against switching from `undefined` to `false` (default value)?

test/resizing.html
```diff
@@ -58,6 +58,16 @@
     </template>
   </test-fixture>
 
+  <test-fixture id="height-by-rows">
+      <template>
```
web-padawan: nit: incorrect indentation here and below

src/vaadin-grid.html
```diff
@@ -389,7 +389,8 @@
           heightByRows: {
             type: Boolean,
             value: false,
-            reflectToAttribute: true
+            reflectToAttribute: true,
```
web-padawan: is the reflecting to attribute related to the fix? I'm wondering whether we need it as an attribute.

samiheikki: I assume it is needed as the attribute is used for styling: https://github.com/vaadin/vaadin-grid/blob/1d8c80a93e5da57f12df09a61f259c085df909d4/src/vaadin-grid-styles.html#L41

src/vaadin-grid.html
samiheikki: Good point, we can avoide one unnecessary notifyResize. Fixed.

test/resizing.html
samiheikki: Fixed.

PR #1558, opened 2019-02-10
Reset keyboard navigation on column hide

### LINKED ISSUE #1129: Unable to move focus from header to body when selection column is hidden
### Description
Focus isn't moving from the header if selection column was included and then hidden. 

recorded

### Expected outcome
Focus is moving to the table body on <kdb>Tab</kbd> press.

### Actual outcome
Focus is **not** moving to the table body on <kdb>Tab</kbd> press.

### Live Demo
https://jsfiddle.net/6r5htd05/84/

### Steps to reproduce
1. Check the checkbox
2. Try to move focus from header to body

samiheikki:
Fixes #1129

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1558)

### DIFF
--- a/src/vaadin-grid-keyboard-navigation-mixin.html
+++ b/src/vaadin-grid-keyboard-navigation-mixin.html
@@ -515,7 +515,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
     _resetKeyboardNavigation() {
       if (this.$.header.firstElementChild) {
-        this._headerFocusable = this.$.header.firstElementChild.firstElementChild;
+        this._headerFocusable = Array.from(this.$.header.firstElementChild.children).filter(el => !el.hidden)[0];
       }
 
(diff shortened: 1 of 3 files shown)


src/vaadin-grid-keyboard-navigation-mixin.html
```diff
@@ -515,7 +515,7 @@
 
     _resetKeyboardNavigation() {
       if (this.$.header.firstElementChild) {
-        this._headerFocusable = this.$.header.firstElementChild.firstElementChild;
+        this._headerFocusable = Array.from(this.$.header.firstElementChild.children).filter(el => !el.hidden)[0];
```
tomivirkki: Prior to filtering, it would be better to sort the columns by `column._order` so that the focusable is the _visually_ first cell in the row. Not a blocked for this PR as is still fixes the reported issue.

PR #1559, opened 2019-02-11
docs: use expenses categories in the tree-grid demos [skip ci]

### LINKED ISSUE #1554: Improve TreeGrid demo data
The demo on vaadin.com is not very useful as it stands. Please change the data so that it makes sense; find data with a natural hierarchy. The same needs to be done for the Java demo, you should preferably use the same sort of data.

web-padawan:
Fixes #1554 

> find data with a natural hierarchy

I found that expense categories could be nice example (as we also have Expense manager demo app).
Note that there is only 1 level of nested items in hierarchy. This should be enough for demo I guess.

Attaching the screenshot of how the updated demo looks like:
demo

@tmattsso please take a look.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1559)


tmattsso:
One level of data is fine, and the example looks suitable as well 👍

PR #1564, opened 2019-02-17
Fixes #1536: cell-focus-changed event

### LINKED ISSUE #1536: Event "cell-activate"
### Description
I am using the `vaadin-grid` in Vaadin Flow and I would like to get informations about the active cell.
I have seen, that the `vaadin-grid` supports the `cell-activate` event (See [vaadin-grid-active-item-mixin.html](https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-active-item-mixin.html)). However, this event, in my opinion, is not reliable yet.
It only seems to fires on click, while you can also activate items using the arrow-keys (see [vaadin-grid-keyboard-navigation-mixin.html](https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-keyboard-navigation-mixin.html)).
Also, it does not use the `EventContext`, which gives information about the row and the column (see [vaadin-grid-event-context-mixin.html](https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-event-context-mixin.html)).
Also, it does not seem to be possible to activate a cell programatically, using Vaadin Flow (Java).

### Expected outcome
The `cell-activate`-Event should fire whenever a new cell is activated (focused).
Also, the `event.details` should use the `EventContext`, which includes the `column` and the `item`.

### Actual outcome
The `cell-activate`-Event is only fired when a cell is clicked, the keyboard navigation does not trigger a `cell-activate`-Event.
Also, the `event.detail` does not include the column, but only the row.

### Steps to reproduce
2. Add a listener to `cell-activate` of the grid and log the event.
3. Click on a cell to see the output. There is no information about the column.
4. Navigate around using the arrow keys. No event is logged.

Peppe:
This WIP PR adds an event to vaadin-grid called `cell-focus-changed`. The event is fired every time a cell is focused, either by clicking on it or by navigating with keyboard. The developer can then for example show the value of the currently selected cell value somewhere else. This can be used for example in spreadsheet kind of applications where the selection of a field is reflected into a function field above the cells.

Example usage:
```javascript
const grid = document.querySelector('vaadin-grid');
const output = document.querySelector('#output')
grid.addEventListener('cell-focus-changed', (event) => {
  output.innerText = event.detail.content;
});
```

There are still some problems and I'd like to hear the thoughts of the Vaadin dev team on them.

1. Is the 'cell-focus-changed' good name for the event?
1. What info should be included in the details of the event? Now it has cell (td element) and content (innerText of content). item would probably be nice as well but I don't know how to get to it from cell.
1. The event has bubbles and composed as 'true'. I don't know if it should - I followed example from another event that does that.
1. Is adding a focus listener on every cell the correct way to go, or might there be performance issues or similar?
1. I was not able to simulate a click in test even with helpers.html `click(element)`. Also I copied some keyboard navigation code from other tests which didn't either work out. Tests do not pass in this WIP PR right now.
1. I'm not sure if this should be in examples or not. It would be nice to have examples on all listeners, but there are none right now except event context. We could rename 'event context' section to 'events' and add a example of this there, but it would be quite lacking in the broader 'events' scope.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1564)


src/vaadin-grid.html
```diff
@@ -445,6 +445,17 @@
         cell.appendChild(slot);
 
         cell._content = cellContent;
+        
+        cell.addEventListener('focus', () => {
+          this.dispatchEvent(new CustomEvent('cell-focus-changed', {
+            bubbles: true,
+            composed: true,
+            detail: {
+              cell: cell,
```
tomivirkki: The cell element is hidden inside the component's shadow root and shouldn't be accessed from outside. So either just dispatch the event directly on the cell element so that contextual information can be obtained with [getEventContext](https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-event-context-mixin.html#L34) and/or fix the `detail` to only include the properties you'd get from `getEventContext`.

tomivirkki:
> * Is the 'cell-focus-changed' good name for the event?

`-changed` suffix is typically reserved for property changed. Would prefer `cell-focus`

> * What info should be included in the details of the event? Now it has cell (td element) and content (innerText of content). item would probably be nice as well but I don't know how to get to it from cell.

Added a comment

> * The event has bubbles and composed as 'true'. I don't know if it should - I followed example from another event that does that.

Could align this with native `focus` event: https://developer.mozilla.org/en-US/docs/Web/Events/focus

> * Is adding a focus listener on every cell the correct way to go, or might there be performance issues or similar?

The logic could be included [here](https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-keyboard-navigation-mixin.html#L452) to avoid this

> * I was not able to simulate a click in test even with helpers.html `click(element)`. Also I copied some keyboard navigation code from other tests which didn't either work out. Tests do not pass in this WIP PR right now.

Testing `focus` is difficult. Keyboard navigation tests have a hacky workaround for this, but it might be easier just to fire the event in the test as custom events.

* I'm not sure if this should be in examples or not. It would be nice to have examples on all listeners, but there are none right now except event context. We could rename 'event context' section to 'events' and add a example of this there, but it would be quite lacking in the broader 'events' scope.

This could be demoed with `getEventContext` also if the event gets fired on the cell instance. Remember to document the new event [like this](https://github.com/vaadin/vaadin-upload/blob/master/src/vaadin-upload.html#L805) so it ends up in the API docs

src/vaadin-grid.html
Peppe: That's a great idea, and it worked great. Much better than what I had. 

The only issue is that getting the actual content of the cell seems hard and I didn't find a straight-forward way of getting it. You can get the clicked column with `event.detail.column.path`, and you can get the item with `event.detail.item`. However you can't do event.detail.item[event.detail.column.path] as path might be "name.first". For that you should do `event.detail.item['name']['path]` or `item.name.path`. Leaving this up to the devs using our code will force them to write a lot of boilerplate and they will run into bugs later when they have taken only into account the single level path like 'firstname' and later add deep paths like 'address.city'.

Am I missing something? Is there an easy way to get the value of the cell from eventContext? is there a function in grid that takes item and column and returns the innerText? Should I add functionality to parse this from item and path, or should I still provide `cell._content.innerText` in the details?

Peppe:
- Event named `cell-focus`.
- event details continued in review discussion.
- `bubbles: true` removed and `composed: true` kept to match the given link.
- Code moved to vaadin-grid-keyboard-navigation-mixin.html.
- I fire now a 'focus' event in the test to see that the code reacts. It worked as long as the event firing for cell-focus was in vaadin-grid.html. When I moved it to the keyboard-navigation-mixin, it stopped working. I tried, "focus", "mouseclick", "click", "mousedown"/"mouseup" and focus(). Nothing seems to work anymore, while the demo still works well. :(
- Demo added to `grid-event-context-demos`
- Event documentation added to vaadin-grid. I feel like a pioneer by adding the first one. I couldn't however find the documentation from the analysis output on http://127.0.0.1:3000/components/vaadin-grid/#/elements/vaadin-grid/

src/vaadin-grid.html
tomivirkki: The problem with providing the inner text is that it's only a visual representation of the data and in practice can be anything. Especially if a developer assigns a renderer or a template to a column, inner text might might produce an unexpected value for the event details property.

tomivirkki: ..for example if a column renderer that localizes the cell content is used, the value of `content` would depend on the user's locale

Peppe: Yes good point. Do you think getting the value from `items` would be good? Do you have a suggestion on how to do that from having the item and the column definition?

tomivirkki: At least Polymer has a helper for that `Polymer.Base.get(item, path)`

Peppe: `Polymer.Base.get(path, item)` is now in use.

demo/grid-event-context-demos.html
```diff
@@ -50,6 +50,41 @@ <h3>Event Context</h3>
...
+          <vaadin-grid-column path="name.last" header="Last name"></vaadin-grid-column>
+          <vaadin-grid-column path="email"></vaadin-grid-column>
+        </vaadin-grid>
+        <div>Clicked cell: <span id="output">None</span></div>
+        <script>
+          window.addDemoReadyListener('#grid-event-context-demos-cell-focus', function(document) {
+            const grid = document.querySelector('vaadin-grid');
+            grid.items = Vaadin.GridDemo.users;
+            const output = document.querySelector('#output');
+            grid.addEventListener('cell-focus', (event) => {
```
tomivirkki: Only ES5 allowed in the demos so that they work with IE11

src/vaadin-grid.html
tomivirkki: Hmm...now the `content` property is available only in events that occur on columns that define a `path`. That's kind of a strange limitation.

When users work with any other type of events, they need to get the context manually. Is there a reason why `cell-focus` should be an exception? To align this event, you'd have something like:
```js
grid.addEventListener('cell-focus', (event) => {
  const context = event.target.getEventContext(event);
  const content = Polymer.Base.get(context.column.path, context.item);
  output.textContent = content;
});
```
...where the event itself wouldn't automatically ship with any contextual data. This would also make it more lightweight – in most cases the data is not even needed. wdyt?

demo/grid-event-context-demos.html
Peppe: I guess the ` (event) => {...}` part was the ES6 part. I changed it to `function(event) {...}`.

moderndeveloperllc:
@Peppe They are moving to Polymer 3, so the your changes will need to moved to the new .js files if you want to keep working on this.

niiyeboah:
I rebased @Peppe's branch and changed [this line](https://github.com/vaadin/vaadin-grid/blob/bb58994e31a9bd96393904d1c2e0443863db871e/src/vaadin-grid-keyboard-navigation-mixin.js#L511) to use `PolymerElement` instead.

tomivirkki:
> ...changed this line to use PolymerElement instead.

Please see the feature design document for updated technical design

web-padawan:
It has been a while since this PR was submitted. Apologies about keeping it open for so long without any progress. 

The component has been converted to Polymer 3 so there are merge conflicts that would be tricky to solve. 
Also, the new development for Vaadin web components will happen in the [monorepo](https://github.com/vaadin/vaadin-web-components). 

I'm closing this now. We will submit a new PR to the monorepo when the linked issue is prioritised.

web-padawan:
Reopened. Please rebase on top of `5.7` branch if this is targeting Vaadin 14 (we will need to create `5.8` branch for it).

Otherwise, if this is for Vaadin 20+ then let's submit a new PR to the monorepo, as mentioned above.

niiyeboah:
Creating replacement PR for [web-components](https://github.com/vaadin/vaadin-web-components).

PR #1568, opened 2019-02-23
Add column-reorder event

### LINKED ISSUE #1537: ColumnReorderListener
### Description
Columns can be reordered in `vaadin-grid` using drag-and-drop.
However, it does not seem to be possible to get this new column-order.

### Expected outcome
The `vaadin-grid` should fire an event when the columns are reordered.
The event should contain the new column order.

### Actual outcome
The columns are reordered but it is not possible (or extremely complex) to get this new order.

### Usecase
I would like to create a configurable grid. The users can reorder and resize the columns to their needs.
I would like to be able to save their current configuration, so that they get the same grid, when they refresh the page.

samiheikki:
Fixes #1537

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1568)

### DIFF
--- a/src/vaadin-grid-column-reordering-mixin.html
+++ b/src/vaadin-grid-column-reordering-mixin.html
@@ -175,6 +175,16 @@ This program is available under Apache License Version 2.0, available at https:/
       this._draggedColumn = null;
       this._lastDragClientX = null;
       this._reorderGhost.style.visibility = 'hidden';
+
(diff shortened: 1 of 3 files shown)


samiheikki:
> The new event should be documented also. Example: https://github.com/vaadin/vaadin-grid/pull/1564/files#diff-fa0c294fc77193488b21e163be2935a0R767

Added documentation.

@tomivirkki should the event actually be `columns-reorder` as there is actually more than one column switching its order?

PR #1595, opened 2019-04-11
feat: column auto-width

### LINKED ISSUE #1233: Column default width should be based on sampled content
The current default width for all columns is a hardcoded `100px`. A better default would be to leave the width undefined and use the content as the measured width, which could be overridden with an explicit width set by the developer or by the end user (column resizing).

The first set/page of data should be used for the measurement, and use the largest width of each column.

I suppose this is even fairly simple to implement – on the first render, the columns are rendered without a width or flex-grow, so they get their natural width. Then measure all the cells in each row (if no explicit width is set for the column) and pick the largest number and set that as the explicit width of the column and restore flex-grow to the previous value.

Haprog:
Fixes #1233

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1595)

### DIFF
--- a/demo/grid-columns-demos.html
+++ b/demo/grid-columns-demos.html
@@ -60,7 +60,7 @@
             grid.items = Vaadin.GridDemo.users;
 
             const button = document.querySelector('vaadin-button');
-            button.addEventListener('click', e => {
+            button.addEventListener('click', function() {
               grid.recalculateColumnWidths();
             });
           });
--- a/demo/grid-columns-demos.html
+++ b/demo/grid-columns-demos.html
@@ -6,6 +6,62 @@
       }
     </style>
 
+    <h3>Automatic Column Width</h3>
+    <p>
+      Applying <code>auto-width</code> on the <code>&lt;vaadin-grid-column&gt;</code> will set the width of the column
+      to automatically fit the content.
+    </p>
+    <p>
+      The column width will be calculated and set only once based on the contents of the initially rendered cells.
+      You can manually trigger the auto sizing behavior again by calling <code>grid.recalculateColumnWidths()</code>.
+    </p>
+    <p>
+      By default the column might still expand if there is more space available, unless you also set <code>flex-grow="0"</code>.
+    </p>
+    <p>
+      <b>Note:</b> If all the columns have <code>flex-grow="0"</code> there might be empty space after the last column
+      if the grid is wider than what is needed to fit all the columns. You probably want to leave at least one column
+      without <code>flex-grow="0"</code> so that column will take all the remaining space that is left.
+    </p>
+    <vaadin-demo-snippet id='grid-columns-demos-auto-width'>
+      <template preserve-content>
+        <vaadin-button>Recalculate column widths</vaadin-button>
+        <vaadin-grid theme="compact column-borders" aria-label="Automatic Column Width Example">
+          <vaadin-grid-column auto-width flex-grow="0" text-align="end" header="#"></vaadin-grid-column>
+          <vaadin-grid-column auto-width flex-grow="0" text-align="center"></vaadin-grid-column>
+          <vaadin-grid-column auto-width flex-grow="0" path="name.first"></vaadin-grid-column>
+          <vaadin-grid-column auto-width flex-grow="0" path="name.last"></vaadin-grid-column>
+          <vaadin-grid-column auto-width flex-grow="0" path="location.city"></vaadin-grid-column>
+          <vaadin-grid-column auto-width flex-grow="0" path="location.state"></vaadin-grid-column>
+          <vaadin-grid-column auto-width path="location.street"></vaadin-grid-column>
+        </vaadin-grid>
+
+        <script>
+          window.addDemoReadyListener('#grid-columns-demos-auto-width', function(document) {
+            const grid = document.querySelector('vaadin-grid');
+            const button = document.querySelector('vaadin-button');
+            grid.items = Vaadin.GridDemo.users;
+            button.addEventListener('click', e => {
+              grid.recalculateColumnWidths();
+            });
+
+            const columns = document.querySelectorAll('vaadin-grid-column');
+            columns[0].renderer = function(root, column, rowData) {
+              root.textContent = rowData.index;
+            };
+
+            columns[1].renderer = function(root, column, rowData) {
+              if (!root.firstElementChild) {
+                root.innerHTML = '<iron-image width="25" height="25" sizing="cover"></iron-image>';
+              }
+              root.firstElementChild.alt = rowData.item.name.first + ' ' + rowData.item.name.last;
+              root.firstElementChild.src = rowData.item.picture.thumbnail;
+            };
+          });
+        </script>
+      </template>
+    </vaadin-demo-snippet>
+
     <h3>Column Grouping</h3>
     <p>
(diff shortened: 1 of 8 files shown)


demo/grid-columns-demos.html
```diff
@@ -6,6 +6,62 @@
...
+      The column width will be calculated and set only once based on the contents of the initially rendered cells.
+      You can manually trigger the auto sizing behavior again by calling <code>grid.recalculateColumnWidths()</code>.
+    </p>
+    <p>
+      By default the column might still expand if there is more space available, unless you also set <code>flex-grow="0"</code>.
+    </p>
+    <p>
+      <b>Note:</b> If all the columns have <code>flex-grow="0"</code> there might be empty space after the last column
+      if the grid is wider than what is needed to fit all the columns. You probably want to leave at least one column
+      without <code>flex-grow="0"</code> so that column will take all the remaining space that is left.
```
tomivirkki: Using `flex-grow="0"` shouldn't probably be promoted as the typical case with this feature. Would rather have only one or two columns use it in the demo.

demo/grid-columns-demos.html
```diff
@@ -6,6 +6,62 @@
...
+    </p>
+    <vaadin-demo-snippet id='grid-columns-demos-auto-width'>
+      <template preserve-content>
+        <vaadin-button>Recalculate column widths</vaadin-button>
+        <vaadin-grid theme="compact column-borders" aria-label="Automatic Column Width Example">
+          <vaadin-grid-column auto-width flex-grow="0" text-align="end" header="#"></vaadin-grid-column>
```
tomivirkki: Using auto-width on this column doesn't look good in the demo as indexes after 99 become clipped and show the ellipsis.

src/vaadin-grid-data-provider-mixin.html
```diff
@@ -326,6 +326,10 @@
 
             this._increasePoolIfNeeded(0);
           }
+
+          if (this._dataProviderDone) {
+            this._dataProviderDone();
```
tomivirkki: At this point we can't be sure if the grid is actually visible: if might be on another view of invisible for other reasons. Essentially the following test (needs cleanup/dry) should also pass:
```
it('should have correct column widths once visible', done => {
  grid = fixture('auto-width');
  grid.hidden = true;
  grid.items = [
    {a: 'fubar', b: 'foo', c: 'foo', d: 'a'},
    {a: 'foo', b: 'foo bar baz', c: 'foo', d: 'bar'},
    {a: 'foo', b: 'foo baz', c: 'foo bar', d: 'baz'},
  ];
  // Allowed margin of measurement to keep the test from failing if there are small differences in rendered text
  // width on different platforms or if there are small changes to styles which affect horizontal margin/padding.
  const delta = 5;

  requestAnimationFrame(() => {
    grid.hidden = false;

    // recalculateColumnWidths() is run in requestAnimationFrame
    requestAnimationFrame(() => {
      expect(parseInt(columns[0].width)).to.be.closeTo(71, delta);
      expect(parseInt(columns[1].width)).to.be.closeTo(114, delta);
      expect(parseInt(columns[2].width)).to.be.closeTo(84, delta);
      expect(parseInt(columns[3].width)).to.be.closeTo(107, delta);

      done();
    });

  });
});
```

Another thing we can't yet know here is if the grid actually has rendered rows (`_effectiveSize`). There's a check above that uses `!this._cache.isLoading()`, so if there's a lazy data provider that's still waiting for data to the second page when the data for the first page has already arrived, there might not be any physical rows in the DOM. This case also needs a test case.

Running something like the following here in the data provider mixin _and_ after "animationend" (grid is actually visible) should work.
```
if (!this._initialColumnWidthSet && Array.from(this.$.items.children).filter(row => row.clientHeight).length) {
  this._initialColumnWidthSet = true;
  this.requestAnimationFrame(() => this.recalculateColumnWidths());
}
```

demo/grid-columns-demos.html
Haprog: I was thinking about setting a fixed width for this to fit those numbers, but I thought that's a good case for testing the "Recalculate column widths" button. But I agree it looks quite ugly. I'll change this column to fixed width. There is still some content which will overflow when you scroll so you can see the effect of the button too.

demo/grid-columns-demos.html
Haprog: I was using `flex-grow="0"` mainly for demonstrative purposes so it's easier to see that the columns can actually fit the content without taking too much space. But I agree now it looks a bit like we're promoting using it with `auto-width`. I'll change the demo to only use it for few columns.

I guess it's still fine to leave this note here?

src/vaadin-grid-data-provider-mixin.html
Haprog: Thanks for the input. I improved the logic and tests a bit. Please take another look.

Haprog: Instead of `filter()` in the suggested `Array.from(this.$.items.children).filter(row => row.clientHeight).length` I wanted to use `find()` to avoid reading `clientHeight` of all rows unnecessarily but stop when it find the first one with a truthy `clientHeight`. But since IE11 doesn't have `Array.prototype.find` I made a small helper method `__hasRenderedRows()` which basically does the same logic in a `for..of` loop returning when it find the first matching row. This also makes the `if` condition more readable as `if (!this._initialColumnWidthsSet && this.__hasRenderedRows()) {`

Haprog: `for..of` doesn't seem to work on `NodeList` in iOS 9 Safari. I'll change it to regular `for` loop.

One test is failing only on iOS 11.3 Safari now. Looking into it.

Haprog: New version pushed. All tests should be passing now (at least passed when manually tested all platforms: Chrome, FF, IE11, Edge, Safari, iOS 9/11/12 Safari). Let's see what Travis says.

demo/grid-columns-demos.html
```diff
@@ -6,6 +6,68 @@
...
+                root.innerHTML = '<iron-image width="25" height="25" sizing="cover"></iron-image>';
+              }
+              root.firstElementChild.alt = rowData.item.name.first + ' ' + rowData.item.name.last;
+              root.firstElementChild.src = rowData.item.picture.thumbnail;
+            };
+
+            grid.items = Vaadin.GridDemo.users;
+
+            const button = document.querySelector('vaadin-button');
+            button.addEventListener('click', e => {
```
tomivirkki: Must use ES5 in the demos to support IE11

Haprog: Ok, will fix and push a commit with [skip ci].

Haprog: Arrow function replaced with normal anonymous function. Demo tested working in IE11 now.

Haprog:
After pushing the separate `[skip ci]` commit for updating the demo, the Travis jobs disappeared from the "checks" section. Here are the latest jobs:
- PR: https://travis-ci.org/vaadin/vaadin-grid/builds/520257069 (all green)
- branch: https://travis-ci.org/vaadin/vaadin-grid/builds/520257062

Haprog:
All green https://travis-ci.org/vaadin/vaadin-grid/builds/520257062

PR #1619, opened 2019-05-10
add getCache; allow clearCache on subCache

### LINKED ISSUE #1618: additional API method related to D&D
First of all, million thanks for the new D&D feature - it is great!

I miss two methods (vaadin-grid-data-provider-mixin) for a use case where a draggable tree grid is in-sync with another view where re-ordering of items also happens: 

- `getCacheItem(item, fn)`: get a item from the cache, event if if is not visible on the grid (e.g. a leaf item on a tree node that has been collapsed). It is different from `getCacheAndIndex(index)` as `index` as it can only find cache currently rendered on the grid.
- `clearSubCache(cache)`: clear the cache for a sub-node only (use case: reordering of items in a tree node that has its cache already, but is collapsed. Next time a user expands it, it should reload a new cache ) 

I'll submit a PR soon.

christophe-g:
This should close #1618 

example (on component containing a tree grid and another draggable view in-sync with the grid): 
```js
onDropIntheView(e) {
 
        const model = e.detail.model;

        this._debounceDrop = Polymer.Debouncer.debounce(
           this._debounceDrop,
           Polymer.Async.timeOut.after(300),
           () => {
              // data representation of model in view and grid are not the same, 
              // hence passing a find function
              this.$.grid.clearCache(model, (a, b) => a.$key === b.key);
             });
      }

```

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-grid/1619)


src/vaadin-grid-data-provider-mixin.html
```diff
@@ -338,22 +365,38 @@
 
     /**
      * Clears the cached pages and reloads data from dataprovider when needed.
+     * When an item is provided, will only clear - an reload - cache containing this item.
+     *
+     * @param  {Object}   item - a model to find
+     * @param  {Function} find - the find function: function(a,b) {return true when a === b}}
```
tomivirkki: I think the find function shouldn't be included in the API. The item instances may not equal by reference but there's already a way to match items by a primitive sub-property: the `itemIdPath` -property. You can internally use the `_itemsEqual` function for matching.

christophe-g: use case is as in the above example, where I have to manage instance matching between two datastores (one for the tree grid, one for the synced view) having different key mechanisms: 
 `this.$.grid.clearCache(modelThatIsDroppedOnTheView, (a, b) => a.$key === b.key);`

tomivirkki: Wouldn't this work (without the matcher function)?
```
this.$.grid.clearCache({$key: modelThatIsDroppedOnTheView.key});
```

christophe-g: a84194097a0c0fc52545401e22ce33c950be3d6c

web-padawan:
It has been a while since this PR was submitted. Apologies about keeping it open for so long without any progress. 

The component has been converted to Polymer 3 so there are merge conflicts that would be tricky to solve. 
Also, the new development for Vaadin web components will happen in the [monorepo](https://github.com/vaadin/vaadin-web-components). 

I'm closing this now. We will submit a new PR to the monorepo when the linked issue is prioritised.

Issue #1655, opened 2019-09-19
Opening multiple row details very slow


moderndeveloperllc:
### Description
Opening row details for large (60+) rows at a time can be very slow

### Expected outcome
Reasonably quick (2+ seconds)

### Actual outcome
Pretty slow (5+ seconds)

https://github.com/vaadin/vaadin-grid/blob/7faabd82b8fd5cbe799efe95d89e859fa3a86dca/src/vaadin-grid-row-details-mixin.html#L133-L135

This appears to be an issue with `_toggleDetailsCell` in the row details mixin. It calls both `row.style.setProperty('padding-bottom', ${cell.offsetHeight}px);` and `requestAnimationFrame(() => this.notifyResize());` for each row opened (lines 130-132) . not only a repaint for each row, but `offsetHeight` (very slow) is getting calculated multiple times for each row. I'm _pretty_ sure that line 133 is redundant as `this.notifyResize()` eventually calls `_updateDetailsCellHeights()`.

I was able to come up with a pretty simple fix that works in my tests, but not in the test suite. The function is getting called in 3 places - `_detailsOpenedItemsChanged`, `_rowDetailsTemplateOrRendererChanged`, and `_updateItem` in `vaadin-grid`. Each call looks to be trying to do different things in different situations, so I'm guessing my fix doesn't work for one of them.

I noticed that if I open rows one at a time, `_detailsOpenedItemsChanged` gets called the first time, but not afterwards. Subsequent rows just go through the `_updateItem` logic. Frustrating to try to debug or fix.

My fix involved
```
_toggleDetailsCell(row, item, requestAnimationFrame) {
<SNIP>
  Polymer.flush();
  if(requestAnimationFrame){
    requestAnimationFrame(() => this.notifyResize());
  }
}
```
and then changing `_detailsOpenedItemsChanged` to 
```
Array.from(this.$.items.children).forEach(row => {
  this._toggleDetailsCell(row, row._item, false);
  this._a11yUpdateRowDetailsOpened(row, this._isDetailsOpened(row._item));
  this._toggleAttribute('details-opened', this._isDetailsOpened(row._item), row);
});
requestAnimationFrame(() => this.notifyResize());
```
This means that the repaint only happens once. Since `vaadin-grid`'s `_updateItem` doesn't touch that parameter, you would have to pass in true to get a repaint:
```
if (this._rowDetailsTemplate || this.rowDetailsRenderer) {
  this._toggleDetailsCell(row, item, true);
[...]

PR #1656, opened 2019-09-20
[Optimization] Delay requestAnimationFrame() to speed large detailsOpenedItems changes

moderndeveloperllc:
Don't resize on every details row if we are looping. We can wait for the next frame render. This will help when a large amount of rows is added to `detailsOpenedItems`. Can get a 200%+ speed increase for 100+ details rows. Further optimization may be possible by trying to minimize the times that `offsetHeight` is called on cells as that causes a synchronous style/layout recalc, but that's beyond my ability.

### DIFF
--- a/src/vaadin-grid-row-details-mixin.html
+++ b/src/vaadin-grid-row-details-mixin.html
@@ -81,7 +81,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
         if (this.detailsOpenedItems.length) {
           Array.from(this.$.items.children).forEach((row, item) => {
-            this._toggleDetailsCell(row, item, false);
+            this._toggleDetailsCell(row, item);
           }, this);
           this._update();
         }
@@ -94,10 +94,11 @@ This program is available under Apache License Version 2.0, available at https:/
         return;
       }
       Array.from(this.$.items.children).forEach(row => {
-        this._toggleDetailsCell(row, row._item, false);
+        this._toggleDetailsCell(row, row._item);
         this._a11yUpdateRowDetailsOpened(row, this._isDetailsOpened(row._item));
         this._toggleAttribute('details-opened', this._isDetailsOpened(row._item), row);
       });
+      this._updateDetailsCellHeights();
       requestAnimationFrame(() => this.notifyResize());
     }
 
@@ -108,7 +109,7 @@ This program is available under Apache License Version 2.0, available at https:/
       this._toggleAttribute('frozen', true, cell);
     }
 
-    _toggleDetailsCell(row, item, requestAnimation) {
+    _toggleDetailsCell(row, item) {
       const cell = row.querySelector('[part~="details-cell"]');
       if (!cell) {
         return;
@@ -133,9 +134,6 @@ This program is available under Apache License Version 2.0, available at https:/
           }
 
           Polymer.flush();
-          if (requestAnimation) {
-            requestAnimationFrame(() => this.notifyResize());
-          }
         }
       }
       if (hiddenChanged) {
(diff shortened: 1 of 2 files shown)


moderndeveloperllc:
FYI, on my code with 101 rows, opening all went from 9 seconds to 2 seconds.

Haprog:
I assume this is for fixing #1655?

moderndeveloperllc:
@Haprog,  yes, forgot to link. As you can see, I did manage to get it to pass the test suite after a bit of redoing.

moderndeveloperllc:
@Haprog Anything I can do to help move this along? It would really help as I have lots of pages with lots of details being opened for lots of items.

tomivirkki:
The change seems to make UX of opening the row details a bit worse on IE11 and Edge. Could this glitch be prevented somehow?

Before:
Kapture 2019-10-17 at 11 32 29
After:
Kapture 2019-10-17 at 11 34 28

moderndeveloperllc:
@tomivirkki I fixed the IE/Edge FOUC and sped up the `_updateDetailsCellHeights()` method. By reading the property, then writing the style, the method was sped up be about 40%. I tested this out on the `grid-row-details-demo.html` page by making the first grid `height-by-rows`, adding a `<vaadin-grid-selection-column>` and then the listener from above. I know that the renderer isn't really optimized for that `height-by-rows`, open-all-details kind of thrashing, but this is my attempt to help in these edge cases.

In my case I have clients with 100+ customers who have lots of orders. I hide the orders in the detailsRow, but have always given them the opportunity to open all customers at once and see all the orders at once. Hence the `height-by-rows` and `grid.detailsOpenedItems = grid.selectedItems` logic.

tomivirkki:
Thanks for the update, I'll take another look this week

tomivirkki:
@moderndeveloperllc It's probably not the changes in `_updateDetailsCellHeights()` that improved the performance. Effectively it does exactly the same as the master revision, but with two iterations. I tried reverting the `_updateDetailsCellHeights()` to the master revision but kept the rest of the changes and the same performance improvements were still there.

It's probably the removed lines in `_toggleDetailsCell` that synchronously measured the row's offsetHeight and added it as the padding-bottom to the row. With the changed revision, the measuring is done asynchronously, which improves performance but causes a glitch (recorded on Chrome with the demo changes you suggested):

Kapture 2019-11-15 at 10 48 19

moderndeveloperllc:
@tomivirkki The `_updateDetailsCellHeights` help comes when you are opening all the rows at once, but you are correct in that its not where the majority of the improvement comes from. I didn't see that stutter when I was testing, though I guess that might be because I'm on a rather new, fast machine.

Do you see any way forward?

tomivirkki:
Maybe the details cell could have a separate `opened` state attribute which only gets applied once the opening is finished (in the end of the requestAnimationFrame). Then in `vaadin-grid-styles.html`:
```
[part~="details-cell"]:not([opened]) {
  visibility: hidden;
}
```

moderndeveloperllc:
Not sure how much I can work on that at the moment, but it might be work looking into.

moderndeveloperllc:
@tomivirkki Any chance of this getting looked at? Rendering seems to be an issue even if the details aren't viewed at the outset. 480 rows with details hidden tool about 30 secs to render on a _fast_ machine. 

ScreenShot 2020-03-20 at 11 00 42 AM

moderndeveloperllc:
I should note that even with this PR's changes, the first render of the 400+ rows is fast, but it seems to throw Chrome's renderer into a fit. It takes about 30 seconds to be able to scroll or do anything on the page. That means this PR's fixes are probably a wash.

tomivirkki:
@moderndeveloperllc I'm spending some time next week on the grid's performance analysis. I'll make sure to include this issue/PR.

moderndeveloperllc:
@tomivirkki Let me know if there is anything I can do to help. On a related not to the details rendering,  I've also noticed is that if the columns have a default sort order, the column renders will go through there cycle before _and_ after the sort happens. Perhaps there may be a way to delay the rendering until after the default sort happens?

moderndeveloperllc:
@tomivirkki Have you had an opportunity to look into this? Should I close out this PR as it looks like the code, as is, won't help, or do you want to keep it active as it's a partial answer for the moment?

tomivirkki:
@moderndeveloperllc I'm currently working on performance tests for the grid (Flow). Added a grid with the row details open as one of the cases; not sure if it covers this use-case properly, though. Would appreciate if you'd take a look and possibly suggest some changes to the test case: https://github.com/vaadin/vaadin-grid-flow/tree/benchmark Once the tests are in place, we'll evaluate different options for optimizing the grid's performance.

moderndeveloperllc:
@tomivirkki I'm not a Java dev, but from what I can parse out, I don't think this case is covered. I see where in integration tests you are checking that when two checkboxes are opened, they both display, but I don't see anywhere in benchmarks where you are testing the selectAll box being checked and detailsOpenedItems == selectedItems. This with a renderer for the details row is the heart of the matter.

moderndeveloperllc:
@tomivirkki Going to close as the move to Polymer 3 is going to invalidate any pulls. I'll see when it moves to beta if this is still an issue.

PR #1665, opened 2019-10-09
fix(scrolling): implementation of WheelEvent deltaMode (#1648)

### LINKED ISSUE #1645: Grid scrolling within Firefox
### Description

I have created a tile view which works perfectly within Chrome but not when using Firefox. The scrolling within Firefox seems to scroll 3 rows for each scroll event. This is an issue as the rows for the tile are responsive and have large heights set to them. This means it will skip row 2 and 3 even if it is not visible to the user. Chrome somehow scrolls by pixels and makes the experience much smoother even for a normal grid. 

Is this a bug or intended for the grid? If it is intended, how would I get Firefox to scroll more like Chrome within my project?

Haprog:
Fix implementation of WheelEvent.DOM_DELTA_LINE to scroll by "lines of
text" instead of by average height of grid rows. This old implementation
caused bad UX especially in cases where the grid rows are very tall.

Implement missing mode WheelEvent.DOM_DELTA_PAGE to scroll by "pages"
instead of pixels. When this was this missing, Firefox on Windows was
scrolling only 1 pixel at a time (super slowly) if the user had set
mouse settings to scroll by one page at a time.

It seems that Firefox is the only browser that implements scrolling with
DOM_DELTA_LINE or DOM_DELTA_PAGE modes depending on OS mouse settings.
On macOS Firefox seems to default to scrolling by 1 line at a time in
DOM_DELTA_LINE mode. On Windows it respects the mouse settings which can
be set to scroll by a selected amount of lines at a time or a single
page at a time. On Windows 10 the default is 3 lines at a time. The
height of a single line is defined as the default font size in the
browser settings. This usually defaults to 16px in modern browsers.

Edge, IE11, Chrome and Safari seem to always use DOM_DELTA_PIXEL mode
instead so this fix doesn't affect them.

Fixes #1645

### DIFF
--- a/src/vaadin-grid-scroll-mixin.html
+++ b/src/vaadin-grid-scroll-mixin.html
@@ -64,7 +64,8 @@ This program is available under Apache License Version 2.0, available at https:/
 
     static get observers() {
       return [
-        '_scrollHeightUpdated(_estScrollHeight)'
+        '_scrollHeightUpdated(_estScrollHeight)',
+        '_scrollViewportHeightUpdated(_viewportHeight)'
       ];
     }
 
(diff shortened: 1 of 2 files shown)


Haprog:
This is a cherry-pick from master. Just checking that all tests pass.

Haprog:
All other tests passed except this one flaky test (which was flaky even before this change) #1666

PR #1677, opened 2019-11-15
Cut down the initial row pool by half (optimize first render)

tomivirkki:
The initial physical row count for the grid is 25 by default. If the individual rows are physically higher or the grid's height is lower than normal, the grid viewport (with padding) could easily be covered with fewer physical rows than 25. This PR reduces the initial row count to 12.


tomivirkki:
This was covered by https://github.com/vaadin/web-components/pull/321

= the grid no longer adds 25 initial rows by default but lets iron-list logic decide when and how many rows are created

PR #1680, opened 2019-11-27
Fix arguments for setDragImage

### LINKED ISSUE #1679: Row drag start causes an exception on iOS 13 iPad Safari
iPad doesn't use the mobile polyfill and https://github.com/vaadin/vaadin-grid/blob/master/src/vaadin-grid-drag-and-drop-mixin.html#L137 without arguments throws

tomivirkki:
Fixes #1679

### DIFF
--- a/src/vaadin-grid-drag-and-drop-mixin.html
+++ b/src/vaadin-grid-drag-and-drop-mixin.html
@@ -134,7 +134,7 @@ This program is available under Apache License Version 2.0, available at https:/
           const rowRect = row.getBoundingClientRect();
           if (!window.ShadyDOM) {
             if (this._ios) {
-              e.dataTransfer.setDragImage(row);
+              e.dataTransfer.setDragImage(row, null, null);
             } else {
               e.dataTransfer.setDragImage(row, e.clientX - rowRect.left, e.clientY - rowRect.top);
             }
--- a/theme/lumo/vaadin-grid-styles.html
+++ b/theme/lumo/vaadin-grid-styles.html
@@ -134,10 +134,13 @@
       }
 
       [part~="row"][dragstart] {
-        /* Add bottom-space to the row so the drag number doesn't get clipped. Needed for IE/Edge */
-        border-bottom: 100px solid transparent;
         z-index: 100 !important;
         opacity: 0.9;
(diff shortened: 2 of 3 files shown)


src/vaadin-grid-drag-and-drop-mixin.html
```diff
@@ -134,7 +134,7 @@
           const rowRect = row.getBoundingClientRect();
           if (!window.ShadyDOM) {
             if (this._ios) {
-              e.dataTransfer.setDragImage(row);
+              e.dataTransfer.setDragImage(row, null, null);
```
tomivirkki: This is enough to fix the exception (of missing arguments)

theme/lumo/vaadin-grid-styles.html
```diff
@@ -134,12 +134,15 @@
...
       [part~="row"][dragstart] {
-        /* Add bottom-space to the row so the drag number doesn't get clipped. Needed for IE/Edge */
-        border-bottom: 100px solid transparent;
         z-index: 100 !important;
         opacity: 0.9;
       }
 
+      #scroller[ie] [part~="row"][dragstart] {
+        /* Add bottom-space to the row so the drag number doesn't get clipped. Needed for IE */
+        border-bottom: 100px solid transparent;
```
tomivirkki: The extra space caused rendering issue on iOS 13 iPad pro Safari, let's just scope it to IE only. Also, evergreen Edge had changed their rendering engine so that the extra space caused the rows 100px below the row being dragged to get included in the drag image as well (scoping this to IE only is the lesser evil)

theme/lumo/vaadin-grid-styles.html
```diff
@@ -157,7 +160,7 @@
         background: var(--lumo-primary-color-50pct);
       }
 
-      #scroller:not([ios]) [part~="row"][dragstart]:not([dragstart=""])::after {
+      #scroller [part~="row"][dragstart]:not([dragstart=""])::after {
```
tomivirkki: The row count number works on iOS 13 iPad pro Safari so we can just re-enable it for all the browsers. No side-effects on iPhone either.

PR #1685, opened 2020-01-22
Fixes: Width of selection column

### LINKED ISSUE #660: Column Reordering for Grid 2.0
Draw inspiration from `<vaadin-grid>` 1.x and the OS X Finder.

MVP:
- When reordering, the header cell of the column being dragged is shown as the "overlay"
- User can reorder only columns on the same column/group level
- Reordering is enabled/disabled for all columns at once
- Frozen columns can be reordered between other frozen columns, but not unfrozen.
- The reordering happens live while dragging (no animation) instead of doing the drop.
- No need for the blue highlight used in 1.x because reordering happens live.
- Mind the keyboard navigation (right + left arrows) after reordering

hongduc-phan:
Add autoWidth property for vaadin-grid-selection-column.

Fixes #660


hongduc-phan:
For the web component it might be problematic as <vaadin-grid-selection-column width="300px"> would no longer work as before. And for WC, we can set auto-width.
So I close this PR

PR #1689, opened 2020-02-04
Remove hidden cells from DOM

### LINKED ISSUE #1676: Hidden columns should not be rendered on the client-side (feature)
On Vaadin 14.0 and 14.1 if a Grid has hidden columns they seem to have a significant performance impact on the client-side rendering. This is a very common strategy for optimising heavy grids where the grid itself has X columns but we only show a subset. By removing the columns from the Grid, performance is improved, however this could/(should?) be an optimisation that the framework already does? 

Tested with: 
Vaadin 14.0.10, 14.0.12, 14.1.beta1
Chrome Version 78.0.3904.97 (Official Build) (64-bit)
MS Edge: 44.17763.771.0 

While browsers matter, the performance hit happens in all browsers as the hidden columns are in the dom.

DiegoCardoso:
Prevent `hidden` columns to render cells and remove/attach cells when `hidden` property is changed.

Fixes #1676


test/keyboard-navigation.html
```diff
@@ -793,13 +793,18 @@
         });
 
         describe('with hidden columns', () => {
-          it('should skip over hidden column with right arrow', () => {
+          it('should skip over hidden column with right arrow', (done) => {
             grid._columnTree[0][1].hidden = true;
-            body.children[0].children[0].focus();
+            requestAnimationFrame(() => {
```
tomivirkki: nit: Rather than relying on raf and making the test async, use `flushGrid(grid)` here after `...hidden=true`. This should work for the other tests also.

src/vaadin-grid.html
```diff
@@ -549,6 +549,7 @@
         if (row.id !== 'outersizer' && row.id !== 'fixedsizer') {
           row.hidden = true;
         }
+        columns = columns.filter(column => !column.hidden);
```
tomivirkki: nit: Instead of reassigning the `column` argument, use
```
columns
  .filter(column => !column.hidden)
  .forEach(...)`

File: test/keyboard-navigation.html (hunk 1 above)
Comment (DiegoCardoso) [reply test/keyboard-navigation.html]: Done. Thanks.

File: src/vaadin-grid.html (hunk 2 above)
Comment (DiegoCardoso) [reply src/vaadin-grid.html]: Done.

File: src/vaadin-grid-column.html [hunk 3]
```diff
@@ -470,9 +470,30 @@
         this.parentElement._columnPropChanged('hidden', hidden);
       }
 
-      this._allCells.forEach(cell => this._toggleAttribute('hidden', hidden, cell));
-
       if (!!hidden !== !!this._previousHidden && this._grid) {
+        if (this._previousHidden && this._cells) {
+          this._cells.forEach(cell => {
+            this._grid.appendChild(cell);
```
tomivirkki: Cells (´<td>´ elements) should never appear as direct children of `<vaadin-grid>`. This loop should probably be removed

test/keyboard-navigation.html
tomivirkki: Should work also in `column.html` tests

src/vaadin-grid-column.html
DiegoCardoso: When I remove this loop, an error is raised at the `__setColumnTemplateOrRenderer` method for the same class, when it calls `this._grid.__getRowModel(cell.parentElement)` and `parentElement` is `null`.  Not sure if I could first test for the `parentElement` and skip.

tomivirkki: It gets invoked after every row update (even if not yet all column cells are added on rows). You could filter out only the cells that have been attached with `cells.filter(cell => cell.parentElement).forEach(cell => {...` or better yet not loop the cells until all of them are attached (more performant)

DiegoCardoso: How could we hold the loop until all cells are attached? 🤔

DiegoCardoso: Moved the logic of handling cells from a column that is moving from `hidden` to visible, so that `_renderColumnTree` of `grid` only notifies the change of cells after the last cell is reattached to the proper parent.

test/keyboard-navigation.html
DiegoCardoso: Removed. Thanks.

src/vaadin-grid-column.html
```diff
@@ -470,9 +470,25 @@
...
       }
 
-      this._allCells.forEach(cell => this._toggleAttribute('hidden', hidden, cell));
-
       if (!!hidden !== !!this._previousHidden && this._grid) {
+        if (hidden === true) {
+          this._allCells.forEach(cell => {
+            if (cell._content.parentNode) {
+              cell._content.parentNode.removeChild(cell._content);
+              cell._vacant = true;
```
tomivirkki: The optimization seems to work well. These lines still need tests though.

DiegoCardoso: Removed setting cell vacant as it is being done at `_rowUpdate` method already. Added test to check children count before and after a column is hidden.

PR #1694, opened 2020-02-19
Rerender grid after loading items. 

tulioag:
- Improves tree grid loading when expand recursively is called (https://github.com/vaadin/vaadin-grid-flow/issues/476) 
- Grid is rerendered after every callback. Undoes #1491 
- ensureSubCacheForScaledIndex is only called for items in this.$.items.children

### DIFF
--- a/src/vaadin-grid-data-provider-mixin.html
+++ b/src/vaadin-grid-data-provider-mixin.html
@@ -231,13 +231,13 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     _isExpanded(item) {
-      return !!this.__expandedKeys.get(this.getItemId(item));
+      return this.__expandedKeys.has(this.getItemId(item));
     }
 
     _expandedItemsChanged(e) {
-      this.__expandedKeys = new Map();
+      this.__expandedKeys = new Set();
       this.expandedItems.forEach(item => {
-        this.__expandedKeys.set(this.getItemId(item), true);
+        this.__expandedKeys.add(this.getItemId(item));
       });
       this._cache.updateSize();
       this._effectiveSize = this._cache.effectiveSize;
--- a/src/vaadin-grid-data-provider-mixin.html
+++ b/src/vaadin-grid-data-provider-mixin.html
@@ -154,6 +154,7 @@ This program is available under Apache License Version 2.0, available at https:/
     static get observers() {
       return [
         '_sizeChanged(size)',
+        '_itemIdPathChanged(itemIdPath)',
         '_expandedItemsChanged(expandedItems.*)'
       ];
     }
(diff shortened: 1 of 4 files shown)


src/vaadin-grid-data-provider-mixin.html
```diff
@@ -295,10 +299,12 @@
           }
 
           // Populate the cache with new items
+          const currentItems = Array.from(this.$.items.children).map(row => row._item);
+
```
tomivirkki: The placement of the above comment is now wrong. The comment should still be right above the forEach loop

src/vaadin-grid-data-provider-mixin.html
```diff
@@ -231,10 +231,14 @@
...
     _isExpanded(item) {
-      return this.expandedItems && this._getItemIndexInArray(item, this.expandedItems) > -1;
+      return !!this.__expandedKeys.get(this.getItemId(item));
     }
 
     _expandedItemsChanged(e) {
+      this.__expandedKeys = new Map();
+      this.expandedItems.forEach(item => {
+        this.__expandedKeys.set(this.getItemId(item), true);
+      });
```
tomivirkki: Wondering if a `Set` would work better (the values in the map are always `true` for each item)

src/vaadin-grid-data-provider-mixin.html
```diff
@@ -231,10 +231,14 @@
...
 
     _isExpanded(item) {
-      return this.expandedItems && this._getItemIndexInArray(item, this.expandedItems) > -1;
+      return this.__expandedKeys.has(this.getItemId(item));
     }
 
     _expandedItemsChanged(e) {
+      this.__expandedKeys = new Set();
+      this.expandedItems.forEach(item => {
+        this.__expandedKeys.add(this.getItemId(item));
```
tomivirkki: If `itemidPath` is changed dynamically, the `__expandedKeys` would expire, but it wouldn't get repopulated. Need to add `itemIdPath` to the observer properties

src/vaadin-grid-data-provider-mixin.html
tulioag: Done

src/vaadin-grid-data-provider-mixin.html
tulioag: Done

src/vaadin-grid-data-provider-mixin.html
tulioag: Created observer that updates `__expandedKeys` if `itemIdPath` changes.

Haprog:
Might be worth noting that this PR also seemed to fix this issue https://github.com/vaadin/vaadin-grid-flow/issues/851

PR #1705, opened 2020-03-25
Calculate column widths on attach

tulioag:
Fixes https://github.com/vaadin/vaadin-grid-flow/issues/906

### DIFF
--- a/test/column-auto-width.html
+++ b/test/column-auto-width.html
@@ -92,7 +92,7 @@
             );
           });
         };
-        grid._recalculateColumnWidths.called = false;
+        sinon.spy.reset();
         whenColumnWidthsCalculated(() => {
           expectColumnWidthsToBeOk(columns);
           done();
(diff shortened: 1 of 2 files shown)


tomivirkki:
Missing a test

test/column-auto-width.html
```diff
@@ -92,7 +92,7 @@
             );
           });
         };
-
+        grid._recalculateColumnWidths.called = false;
```
tomivirkki: Rather use `spy.reset()`

tomivirkki:
Found a synchronous solution for this. We could recalculate the column widths whenever the column set changes:

```js
_columnTreeChanged(columnTree, splices) {
  ...
  this.recalculateColumnWidths();
}
```

test/column-auto-width.html
```diff
@@ -66,7 +67,7 @@
 
       beforeEach(done => {
         grid = fixture('auto-width');
-        sinon.spy(grid, '_recalculateColumnWidths');
+        spy = sinon.spy(grid, '_recalculateColumnWidths');
```
tomivirkki: nit: `grid._recalculateColumnWidths.reset()` would have probably worked fine as well

Issue #1490, opened 2018-11-13
Fix RTL issues


tomivirkki:
- Frozen translate direction
- Sort arrows should have space on the other side also
- Multi-sort indicator is on the wrong side
- Tree toggle arrow faces wrong direction
- Column borders variant places the borders on the wrong side

JonUihlein:
Hello. Apologies if this is the wrong place to report additional RTL inconsistencies, but I believe there is also an issue with `<vaadin-grid-selection-column>` when setting `frozen` to `true` when direction is RTL. The column seems to be hidden behind other columns. Could be an issue with frozen columns in general.

PR #1706, opened 2020-03-25
Initial rtl support

yuriy-fix:
Fixes https://github.com/vaadin/vaadin-grid/issues/1490


src/vaadin-grid-scroll-mixin.html
```diff
@@ -386,14 +386,20 @@
...
     }
 
-    _getTranslate(x, y) {
-      return 'translate(' + x + 'px,' + y + 'px)';
+    _getTranslate(x, y, rtl) {
+      const outerScroller = this.$.outerscroller;
+      const isRTL = this.getAttribute('dir') === 'rtl';
+
+      if (isRTL && rtl) {
+        x = - (outerScroller.scrollWidth - (x + outerScroller.clientWidth));
```
tomivirkki: Considering how often this gets invoked, getting the "dir" attribute from the dom element and calculating `(outerScroller.scrollWidth - outerScroller. clientWidth)` every time might cause some performance impact. Also, instead of adding a boolean parameter to the function, would be cleaner to leave the function as it is and add to the value that is passed to the function as x on line 389.

yuriy-fix: Yeah, in any case the calculation will most probably have performance impact. I am still searching for the proper place (i.e. overriding getters and calculations from `iron-scroll-target-behavior` with `rtl` ), it's just a "working fallback".

theme/lumo/vaadin-grid-tree-toggle-styles.html
```diff
@@ -67,6 +67,31 @@
...
+
+      /* RTL specific styles */
+
+      :host([dir="rtl"]) {
+        margin-left: 0;
+        margin-right: calc(var(--lumo-space-s) * -1);
+      }
+
+      :host([dir="rtl"]) [part="toggle"] {
+        margin-left: 0;
```
DiegoCardoso, line 79: `margin-right` is not reset here. It could be reset or `margin-right` could be changed inside a `:host(:not([dir="rtl"])) [part="toggle"]` selector.

src/vaadin-grid.html
```diff
@@ -770,9 +777,19 @@
       _resizeHandler() {
         this._updateDetailsCellHeights();
         this._accessIronListAPI(super._resizeHandler, true);
+        this._updateScrollerMeasurements();
         this._updateHeaderFooterMetrics();
       }
 
+      _updateScrollerMeasurements() {
```
DiegoCardoso, line 784: Shouldn't this be at `vaadin-grid-scroll-mixin` as it is also using this method?

DiegoCardoso, line 784: Shouldn't this method be at `vaadin-grid-scroll-mixin`, as it is also using it there?

theme/lumo/vaadin-grid-tree-toggle-styles.html
DiegoCardoso, line 79: `margin-right` is not reset here. It could be reset or `margin-right` could be changed inside a `:host(:not([dir="rtl"])) [part="toggle"]` selector.

yuriy-fix: The default value of `margin` is 0. I am just ensuring it with this line for the left side. 
No need in reseting it for the `right` side.

src/vaadin-grid.html
yuriy-fix: It's called within `_resizeHandler` right before this method. I've decided that it would be easier to declare it over here, but can be changed.

theme/lumo/vaadin-grid-tree-toggle-styles.html
DiegoCardoso: Isn't `margin` being set here:
https://github.com/vaadin/vaadin-grid/pull/1706/files#diff-7c687aff3e311564441142d7b52e6161R28-R29

yuriy-fix: Good catch! Updated with `:not([dir="rtl"])`

src/vaadin-grid.html
tomivirkki: If the cached object is only accessed in `vaadin-grid-scroll-mixin` it would be cleaner to place it there to have the code properly organized.

Have you verified that caching this actually optimizes performance? if not, then it's premature and shouldn't be done. There are numerous occasions when scroll width of the grid might change and since we'd need to have this cache always in sync, `_updateScrollerMeasurements` should be invoked in all those places.

DiegoCardoso: hmm.. good point. If they are only being used at `_translateStationaryElements`, it might make sense to use the `outerscroller` values straight from there.

yuriy-fix: I have moved `_updateScrollerMeasurements` to the `vaadin-grid-scroll-mixin` - https://github.com/vaadin/vaadin-grid/pull/1706/commits/1253f93e54455be1f178d83284c5bc530637b709

As for the optimization for the caching, I have verified it with the number of calls to calculate value out from `outerScroller.scrollWidth` and `outerScroller.clientWidth` (which is currently done in `_updateScrollerMeasurements`). 

If we are going to do it in `_translateStationaryElements` the amount of those calls will be approximately doubled as it is called from `_scrollHandler` -> `_afterScroll`. 

For the cases I have checked the `_resizeHandler` (where `_updateScrollerMeasurements` is called) is covering the occasions when scroll width of the grid is changing and we need to sync.

The optimization of this behavior is coming from the caution note mentioned in the comment above (https://github.com/vaadin/vaadin-grid/pull/1706#discussion_r400191427). It's not removing all of the calls, but still reduces those amount. Can get rid out of it in order to avoid any possible issues with changes of scroll width as anyway the performance is affected by extra calculations compared to `ltr` mode.

tomivirkki: My earlier statement: "...might cause some performance impact" meant that in theory there might be some performance hit with obtaining scrollWidth on every scroll event. But it was still hypothetical, not a fact, so a verification would have been preferred before optimizing. I think it's ok to go with this suggested approach as the optimization is quite trivial.

yuriy-fix:
Small `DnD` note. In the current implementation the `count` position is still displayed on the right side of the pointer. Is that correct, @mehdi-vaadin?

Compare two screenshots below with the number of dragged rows in red box (`ltr` and `rtl` mode):
Screen Shot 2020-04-06 at 14 47 10
Screen Shot 2020-04-06 at 14 46 18

test/helpers.html
```diff
@@ -205,6 +205,22 @@
...
 
+  window.isWithinParentConstraints = (el, parent) => {
+    ['top', 'bottom', 'left', 'right'].forEach(constraint => {
+      if (el.getBoundingClientRect[constraint] !== parent.getBoundingClientRect[constraint]) {
+        return false;
+      }
+    });
+
+    return true;
+  };
```
DiegoCardoso, line 216: This function will always return `true` is the `return false` statement is within the `forEach` callback. Maybe using `every` or `some` would make more sense here

bower.json
```diff
@@ -64,7 +64,7 @@
     "iron-ajax": "PolymerElements/iron-ajax#^v2.0.0",
     "app-localize-behavior": "^v2.0.0",
     "iron-list": "^v2.0.4",
-    "vaadin-demo-helpers": "vaadin/vaadin-demo-helpers#^3.0.0",
+    "vaadin-demo-helpers": "vaadin/vaadin-demo-helpers#^3.1.0-alpha1",
     "vaadin-button": "vaadin/vaadin-button#^2.1.0",
     "vaadin-context-menu": "^4.3.1"
```
DiegoCardoso, line 69: nit: can be updated.

test/frozen-columns.html
```diff
@@ -36,135 +36,151 @@
...
+        let grid;
+        let columns;
+        let isRTL;
+
+        beforeEach(() => {
+          grid = fixture('frozen-columns');
+          grid.dataProvider = infiniteDataProvider;
+          columns = grid.querySelectorAll('vaadin-grid-column');
+          document.documentElement.setAttribute('dir', direction);
+          isRTL = direction === 'rtl';
```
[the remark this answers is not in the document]
yuriy-fix: Done.

bower.json
yuriy-fix: Done.

mehdi-vaadin:
> Small `DnD` note. In the current implementation the `count` position is still displayed on the right side of the pointer. Is that correct, @mehdi-vaadin?

IMO, it's totally ok for the `count` to be on the right. It's more related to the direction of the mouse pointer rather than the direction of the component/page.

PR #1722, opened 2020-04-29
perf: skip scrollToIndex(0) on _effectiveSize changes

pekam:
`scrollToIndex` causes two calls to costly `_assignModels`, even when nothing actually changed.

This causes a lot of extra computation, especially with the Flow component method `TreeGrid::expandRecursively`, which repeatedly changes `_effectiveSize`, which in turn resets the scroll position.

Notice how `_effectiveSizeChanged` calls `_assignModels` three times, two of which are coming from `scrollToIndex`:

image

The initial approach was to check `if (scrollTop === 0 && index === 0)` in
the `scrollToIndex` function, but this caused many test failures, as
apparently grid depends on the side-effects of this function in some
situations.

Related to vaadin/vaadin-grid-flow#891

### DIFF
--- a/src/vaadin-grid-scroller.html
+++ b/src/vaadin-grid-scroller.html
@@ -131,7 +131,9 @@ This program is available under Apache License Version 2.0, available at https:/
         this._virtualCount = Math.min(this.items.length, size) || 0;
 
         if (this._scrollTop === 0) {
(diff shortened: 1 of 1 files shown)


Haprog:
This had to be reverted. See https://github.com/vaadin/vaadin-grid/releases/tag/v5.6.1

lkraav:
Is this performance win coming back in a rearchitected way?

PR #1731, opened 2020-05-06
Update scroller measurements on dir changed

### LINKED ISSUE #1726: RTL: Cannot read property 'clientWidth' of undefined
The following test is failing in Travis:

```
chrome 81                ✓ test/keyboard-navigation.html » keyboard navigation » navigating with keys » navigation left / right in rtl » should navigate on left when navigation mode is off
chrome 81                Tests failed: Error thrown outside of test function: Cannot read property 'clientWidth' of undefined
  HTMLElement._translateStationaryElements at /components/[secure]/src/[secure]-scroll-mixin.html.js:364
                  HTMLElement._afterScroll at /components/[secure]/src/[secure]-scroll-mixin.html.js:223
                HTMLElement._scrollHandler at /components/[secure]/src/[secure]-scroller.html.js:289
```

yuriy-fix:
Fixes #1726

### DIFF
--- a/src/vaadin-grid.html
+++ b/src/vaadin-grid.html
@@ -427,6 +427,7 @@ This program is available under Apache License Version 2.0, available at https:/
         super.attributeChangedCallback(name, oldValue, newValue);
         if (name === 'dir') {
           this.__isRTL = newValue === 'rtl';
+          this._updateScrollerMeasurements && this._updateScrollerMeasurements();
         }
       }
 
--- a/src/vaadin-grid.html
+++ b/src/vaadin-grid.html
@@ -427,7 +427,7 @@ This program is available under Apache License Version 2.0, available at https:/
         super.attributeChangedCallback(name, oldValue, newValue);
         if (name === 'dir') {
           this.__isRTL = newValue === 'rtl';
-          this._updateScrollerMeasurements && this._updateScrollerMeasurements();
+          this._updateScrollerMeasurements();
         }
       }
 
--- a/test/frozen-columns.html
+++ b/test/frozen-columns.html
@@ -51,6 +51,20 @@
           flushGrid(grid);
         });
 
+        after(() => document.documentElement.removeAttribute('dir'));
+
+        !isRTL && it('should update scrollerMetrics on dir changed to rtl after frozen columns set', done => {
+          columns[0].frozen = true;
+          grid._debouncerCacheElements.flush();
+          grid.setAttribute('dir', 'rtl');
+          listenOnce(grid.$.table, 'scroll', () => {
+            expect(grid.__scrollerMetrics).not.to.be.undefined;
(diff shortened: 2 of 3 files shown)


src/vaadin-grid.html
```diff
@@ -427,6 +427,7 @@
         super.attributeChangedCallback(name, oldValue, newValue);
         if (name === 'dir') {
           this.__isRTL = newValue === 'rtl';
+          this._updateScrollerMeasurements && this._updateScrollerMeasurements();
```
tomivirkki, line 430: Why the check for `_updateScrollerMeasurements`? Is there a chance it doesn't exist (it's defined in scroll-mixin)?

web-padawan, line 430:
```suggestion
          this._updateScrollerMeasurements();
```

Not sure why we need this?

test/frozen-columns.html
```diff
@@ -51,6 +51,20 @@
...
         });
 
+        after(() => document.documentElement.removeAttribute('dir'));
+
+        !isRTL && it('should update scrollerMetrics on dir changed to rtl after frozen columns set', done => {
+          columns[0].frozen = true;
+          grid._debouncerCacheElements.flush();
+          grid.setAttribute('dir', 'rtl');
+          listenOnce(grid.$.table, 'scroll', () => {
+            expect(grid.__scrollerMetrics).not.to.be.undefined;
```
tomivirkki, line 61: This checks for a very specific internal detail. Isn't there a black-box way to test the fix? Something like `expect(scrollTheGridFunction).not.to.throw(Error)`

yuriy-fix: I have initially tried to go with `black-box` approach (both with `not.to.throw(Error)` and `spy`), but it doesn't seem to work properly. `__setNormalizedScrollLeft` doesn't throw.

I have an idea with:

```
expect(() => grid._scrollHandler()).not.to.throw(Error);
```

But it's a bit of hack as scrolling is not really happening and it's just testing handler called manually which seems to be specific internal detail.

I can try to experiment more and come up with better one.

src/vaadin-grid.html
yuriy-fix: Discussed with Serhii as well. Removed now.

test/frozen-columns.html
tomivirkki: If it's not easily doable, not a blocker for this PR. `_scrollHandler` should get called (synchronously) if you fire a `wheel` event: https://github.com/vaadin/vaadin-grid/blob/67d6d3162a2b699a2fa6a5f2f059f075266db3c5/test/helpers.html#L111

yuriy-fix: Thanks for the hint! Have used wheel event as an emulation of scrolling :)

PR #1735, opened 2020-05-12
Don't run renderers for missing header/footer cells

### LINKED ISSUE #1729: Hidden column causing exceptions since 5.6.0
### Description

Receiving errors since 5.6.0

```bash
Uncaught TypeError: Cannot read property '_content' of undefined
    at HTMLElement.ColumnBaseMixin.__runRenderer (vaadin-grid-column.js:242)
    at HTMLElement.ColumnBaseMixin._renderHeaderAndFooter (vaadin-grid-column.js:235)
    at eval (vaadin-grid.js:771)
    at Array.forEach (<anonymous>)
    at eval (vaadin-grid.js:771)
    at Array.forEach (<anonymous>)
    at HTMLElement.GridElement.render (vaadin-grid.js:770)
```

### Live Demo

https://stackblitz.com/edit/typescript-ybspyn

### Steps to reproduce

1. Have a column start out as `hidden`
2. Manually run `grid.render()`

In the repro above, I had made the connection to rendering ... but the repro is over-complicated - it copied a v-grid selection demo because that was closer to my production code.  It appears all that is necessary is just running `render`

grid-5 6 0-hidden-col

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE 11
- [x] iOS Safari
- [x] Android Chrome

tomivirkki:
Fixes #1729

### DIFF
--- a/src/vaadin-grid-column.html
+++ b/src/vaadin-grid-column.html
@@ -235,10 +235,10 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     _renderHeaderAndFooter() {
-      if (this.headerRenderer) {
+      if (this.headerRenderer && this._headerCell) {
         this.__runRenderer(this.headerRenderer, this._headerCell);
       }
-      if (this.footerRenderer) {
+      if (this.footerRenderer && this._footerCell) {
         this.__runRenderer(this.footerRenderer, this._footerCell);
       }
     }
--- a/test/column.html
+++ b/test/column.html
@@ -182,11 +182,21 @@
             expect(spy.called).to.be.false;
           });
 
-          it('should dettach cells from a hidden column', () => {
+          it('should detach cells from a hidden column', () => {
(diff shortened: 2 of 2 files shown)


src/vaadin-grid-column.html
```diff
@@ -235,10 +235,10 @@
     }
 
     _renderHeaderAndFooter() {
-      if (this.headerRenderer) {
+      if (this.headerRenderer && this._headerCell) {
```
tomivirkki, line 238: Also considered having the check in
- vaadin-grid's `render()` but from the grid's point of view, calling `_renderHeaderAndFooter` should be safe regardless of the column's state.
- vaadin-grid-column's `__runRenderer()` but the function may expect the `cell` argument to be defined; it's the responsibility of the caller to make sure the arguments are valid. Otherwise, it wouldn't make sense to have a check for `this.headerRenderer` here either.

PR #1739, opened 2020-05-27
Click first cell child on space keyup

yuriy-fix:
Fixes https://github.com/vaadin/vaadin-confirm-dialog-flow/issues/120

**Important note**: Changes the grid behavior by clicking the cell on `Space` keyup. However cell is still "activated" on `Space` keydown.

### DIFF
--- a/test/keyboard-navigation.html
+++ b/test/keyboard-navigation.html
@@ -185,8 +185,12 @@
       MockInteractions.keyDownOn(target || grid.shadowRoot.activeElement, 40, [], 'ArrowDown');
     }
 
-    function space(target) {
+    function spaceDown(target) {
       MockInteractions.keyDownOn(target || grid.shadowRoot.activeElement, 32, [], ' ');
+    }
+
+    function spaceUp(target) {
+      MockInteractions.keyUpOn(target || grid.shadowRoot.activeElement, 32, [], ' ');
     }
 
     function pageUp(target) {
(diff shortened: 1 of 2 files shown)


test/keyboard-navigation.html
```diff
@@ -1416,20 +1420,22 @@
...
+          it('should click first cell child on space keyup', () => {
             const firstChild = getCellContent(header.children[0].children[2]).children[0];
             const clickStub = sinon.stub(firstChild, 'click');
 
-            space();
+            spaceDown();
+            spaceUp();
 
             expect(clickStub.called).to.be.true;
           });
```
pekam, line 1431: To verify that the bug is fixed, there should be also a test case for calling only `spaceDown()`, expecting the element not to be clicked.

yuriy-fix: Good point! I have enhanced the existing test.

knoobie: I'm wondering - could this be the root issue of https://github.com/vaadin/flow/issues/6929 as well?

yuriy-fix:
I was thinking about that, but explicitly preserved the `activeItem` behavior in order to avoid breaking changes to existing applications.

Active-item behavior is working this way on `master` and probably somebody relies on it.
I guess we can discuss changing it and create another issue if needed.

pekam:
Yeah, I'm also a bit afraid to make this kind of changes...

But I do think that the case is pretty much the same: To trigger an action, whether that is through a button's click listener or `active-item-changed`, it should happen consistently on key up.

Thinking about the bug ticket, it might be also a valid use case that you want to open a confirm dialog for some action when "activating" the item anywhere in the row, instead of only via the button in one column. But if you'd open the confirm dialog on `active-item-changed` event, you would stumble upon this bug again.

I also think that if changing the `active-item-changed` to trigger on `keydown` is considered a breaking change, then the same is true for this `click` behavior. It's just maybe not as risky.

But I'm not sure about my personal opinion, whether this should be changed now or not...

@tomivirkki, I would like to hear your opinion on this, or even more preferably the absolute truth. ;)

test/keyboard-navigation.html
pekam: Knoobie, good point, it looks related. But that issue seems to happen only with the Flow integration and component renderers, so probably not directly related to this.

Yuriy, tests ok :+1:

PR #1772, opened 2020-07-14
Fix filtering example

### LINKED ISSUE #1770: Filtering doesn't scroll up to the filtered items in the dataprovider demo
On the dataprovider demo: https://vaadin.com/components/vaadin-grid/html-examples/grid-filtering-demos

If you scrolled down in the grid and filter then the user is still at the end on the grid
vaadin_grid_dataprovider_after_filtering

anasmi:
- If there is a new filter with a length longer than 0, index should start from 0 (Previous page size should not have an effect)
- On any change in DataProvider size scroll to the top

Currently,filtering demo for lazy loaded items throws 505, when scrollbar is positioned in the bottom of the grid.

Fixes #1770

### DIFF
--- a/demo/grid-filtering-demos.html
+++ b/demo/grid-filtering-demos.html
@@ -81,9 +81,16 @@
                 grid.size = response.size;
 
                 callback(response.result);
+                grid.scrollToIndex(0);
               };
 
-              const index = params.page * params.pageSize;
+              let index = params.page * params.pageSize;
+              params.filters.forEach(item => {
+                if (item.value.length > 0) {
+                  index = 0;
+                }
+              });
+
               let url = 'https://demo.vaadin.com/demo-data/1.0/people?index=' + index + '&count=' + params.pageSize;
 
               // `params.filters` format: [{path: 'lastName', direction: 'asc'}, ...];


demo/grid-filtering-demos.html
```diff
@@ -81,9 +81,16 @@ <h3>Filtering with Data Provider</h3>
                 grid.size = response.size;
 
                 callback(response.result);
+                grid.scrollToIndex(0);
```
web-padawan, line 84: This causes grid to jump back to top when another request is completed while scrolling:

jump

demo/grid-filtering-demos.html
```diff
@@ -81,9 +81,16 @@ <h3>Filtering with Data Provider</h3>
...
               };
 
-              const index = params.page * params.pageSize;
+              let index = params.page * params.pageSize;
+              params.filters.forEach(item => {
+                if (item.value.length > 0) {
+                  index = 0;
+                }
+              });
```
web-padawan, line 92: nit: using `some` method which would be a bit cleaner

```js
if (params.filters.some(filter => filter.value.length > 0)) {
  index = 0;
}
```

PR #1777, opened 2020-07-20
Fix Javascript error when filtering grid with a large dataset 

jcgueriaud1:
Here is a change for the grid to reset the _vidxOffset when the _effectiveSize is less than virtual count.

It should fix this ticket:
https://github.com/vaadin/vaadin-grid-flow/issues/1053

### DIFF
--- a/src/vaadin-grid-scroller.html
+++ b/src/vaadin-grid-scroller.html
@@ -270,6 +270,8 @@ This program is available under Apache License Version 2.0, available at https:/
         const oldOffset = this._vidxOffset;
         if (this._accessIronListAPI(() => this._maxScrollTop) && this._virtualCount < this._effectiveSize) {
           this._adjustVirtualIndexOffset(delta);
+        } else {
+            this._vidxOffset = 0;
         }
         if (this._vidxOffset !== oldOffset) {
           this._update();
--- a/src/vaadin-grid-scroller.html
+++ b/src/vaadin-grid-scroller.html
@@ -271,7 +271,7 @@ This program is available under Apache License Version 2.0, available at https:/
         if (this._accessIronListAPI(() => this._maxScrollTop) && this._virtualCount < this._effectiveSize) {
           this._adjustVirtualIndexOffset(delta);
         } else {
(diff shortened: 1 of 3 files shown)


jcgueriaud1:
Any idea why travis build failed?

src/vaadin-grid-scroller.html
```diff
@@ -270,6 +270,8 @@
         const oldOffset = this._vidxOffset;
         if (this._accessIronListAPI(() => this._maxScrollTop) && this._virtualCount < this._effectiveSize) {
           this._adjustVirtualIndexOffset(delta);
+        } else {
+            this._vidxOffset = 0;
```
web-padawan, line 274: Please change the indentation here, should be two spaces.

test/filtering-huge-grid.html
```diff
@@ -0,0 +1,122 @@
...
+            var response = [];
+            for (var i = 0; i <= params.pageSize; i++) {
+              var element = params.page * params.pageSize + i;
+              if (element < grid.size) {
+                response.push({value: 'item' + element});
+              }
+            }
+            callback(response);
+          };
+          expect(getCellContent(getFirstVisibleItem(grid)).textContent).to.contain('item0');
```
web-padawan, line 94: Why is `expect` used outside of the test below, in `beforeEach` block?

When I remove this line, I'm getting the following error:

```
Error: expected 'item0' to include 'item299975'
```

src/vaadin-grid-scroller.html
jcgueriaud1: Should be ok now.

test/filtering-huge-grid.html
jcgueriaud1: I removed it and changed the test. It was wrong (and causes the timeout).

web-padawan:
I will to merge this into other branch to trigger SauceLabs tests in all browsers. Then we can continue from there.

jcgueriaud1:
Ok. It's better to run in all browsers :)

PR #1782, opened 2020-08-12
perf: Debounce effective size update

tomivirkki:
The change makes the grid to not update its `_effectiveSize` on every data provider callback but defer the change until all the requests have finished.

This change together with https://github.com/vaadin/vaadin-grid-flow/pull/1084 should have a positive impact especially on TreeGrid performance.

Related to https://github.com/vaadin/vaadin-grid-flow/issues/802


tomivirkki:
As such, this would re-introduce https://github.com/vaadin/vaadin-grid-flow/issues/851 (data requests are left pending by the connector), so it needs to be resolved first

PR #2075, opened 2020-11-12
fix: scroll to added item (#2067) (cherry-pick)

yuriy-fix:
* Add test and fix scroll to

* Remove FF workaround, increase pool on sizeChanged


web-padawan:
The `5.4` branch is only used by Vaadin 14.0 which is no longer LTS and we recommend upgrading to Vaadin 14.4

I merged #2077 which backported the fix for the current Vaadin 14 LTS. Closing.

PR #2076, opened 2020-11-12
fix: scroll to added item (#2067) (cherry-pick)

yuriy-fix:
* Add test and fix scroll to

* Remove FF workaround, increase pool on sizeChanged


tomivirkki (approved):

web-padawan:
The `5.5` branch is used by Vaadin 14.1 and 14.2, these are no longer LTS and we recommend upgrading to Vaadin 14.4

I merged #2077 which backported the fix for the current Vaadin 14 LTS and it will be released as 5.6.x. Closing.

Issue #1683, opened 2020-01-07
Extract items array data provider as a separate module


tomivirkki:
Remove the current `items` array API from the core `<vaadin-grid>` and provide it as a separate module. The motivation for this is to have a clear separation between the core data API (`dataProvider`) and the `items` abstraction built on top of it.

Benefits:
- Clean up the grid core API layer of features that only work with `items` array
- Enable creating richer APIs for the array data provider (without bloating the grid core with even more APIs that only work when `items` is in use)
  - https://github.com/vaadin/vaadin-grid/issues/1125
  - https://github.com/vaadin/vaadin-grid/issues/1142
  - https://github.com/vaadin/vaadin-grid/issues/1154
  - https://github.com/vaadin/vaadin-grid/issues/1526
- Reduce the bundle size impact of `<vaadin-grid>` when `dataProvider` only is used

(very early) API suggestion:
```js
import { arrayDataProvider } from '@vaadin/vaadin-grid/arrayDataProvider'

const data = arrayDataProvider(items);

// Configure all array data provider -related settings

// Item's "posts" property is an array of its child items
data.itemChildArrayPath = 'posts';

// Custom starts-with filter
data.itemFilter = (item, index, filters) => {
  const filter = filters[0];
  const filterValue = filter.value;
  const itemPropertyValue = item[filters.path];
  return itemPropertyValue.startsWith(itemPropertyValue);
};

...

// Finally, set the arrayDataProvider as the grid's dataProvider
grid.dataProvider = data;
```

PR #2083, opened 2020-11-18
Give user choice to supply a custom sorting function

Eis-D-Z:
This aims to allow a user to use his own custom comparison functions to sort the data in a vaadin-grid. The function should be supplied to <vaadin-grid-sorter> eg:
```
<vaadin-checkbox>Enable Multi-Sorting</vaadin-checkbox>

<vaadin-grid aria-label="Sorting Example">
  <vaadin-grid-column path="name.first" id="firstnamecolumn"></vaadin-grid-column>
  <vaadin-grid-column path="name.last" id="lastnamecolumn"></vaadin-grid-column>
  <vaadin-grid-column path="email"></vaadin-grid-column>
</vaadin-grid>

<script>
   function customSort(a, b) {
      // example for case insensitive sorting
      // a and b are instances of elements from items array
       return a.name.first.localeCompare(b.name.first, 'en', {sensitivity:"base"});
   }
  customElements.whenDefined('vaadin-grid').then(function() {
    const grid = document.querySelector('vaadin-grid');
    grid.items = Vaadin.GridDemo.users;

    const checkBox = document.querySelector('vaadin-checkbox');
    checkBox.addEventListener('checked-changed', function(event) {
      grid.multiSort = event.detail.value;
    });

    grid.querySelector('#firstnamecolumn').headerRenderer = function(root) {
      root.innerHTML = '<vaadin-grid-sorter path="name.first" cmp="customSort">First name</vaadin-grid-sorter>';
    };

    grid.querySelector('#lastnamecolumn').headerRenderer = function(root) {
      root.innerHTML = '<vaadin-grid-sorter path="name.last" direction="asc">Last name</vaadin-grid-sorter>';
    };
  });
</script>
```

### DIFF
--- a/src/vaadin-grid-array-data-provider-mixin.html
+++ b/src/vaadin-grid-array-data-provider-mixin.html
@@ -119,9 +119,9 @@ This program is available under Apache License Version 2.0, available at https:/
     _multiSort(a, b) {
       return this._sorters.map(sort => {
         if (sort.direction === 'asc') {
-          return this._compare(Polymer.Base.get(sort.path, a), Polymer.Base.get(sort.path, b));
+          return sort.cmp ? sort.cmp(a, b) : this._compare(Base.get(sort.path, a), Base.get(sort.path, b));
         } else if (sort.direction === 'desc') {
-          return this._compare(Polymer.Base.get(sort.path, b), Polymer.Base.get(sort.path, a));
+          return sort.cmp ? sort.cmp (b, a) : this._compare(Base.get(sort.path, b), Base.get(sort.path, a));
         }
         return 0;
       }).reduce((p, n) => {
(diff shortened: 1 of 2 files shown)


tomivirkki:
Hi
Thanks for the PR. If you need custom sorting, you should use a custom data provider for it for now. The `array-data-provider-mixin` is just one data provider implementation and we don't intend to add more features to it but rather separate it as its own module (see https://github.com/vaadin/vaadin-grid/issues/1683). 
Note that `array-data-provider-mixin` actually has a design flaw as it is in the master branch: it shouldn't access the `<vaadin-grid-sorter>` elements at all but read the sort orders from the `opts.sortOrders` instead. Closing.

PR #2080, opened 2020-11-16
Convert to Polymer 3, add events type definitions

web-padawan:
Fixes #2074

PR #2084, opened 2020-11-20
ci: support skipping, ignore master [skip ci]

web-padawan:
Looks like there are duplicate builds on master when pushing directly with `magi-cli`. Let's fix that.

Any code merged to master should be green anyways, so we don't need to run build twice.


web-padawan:
Closing, will add this to #2080 as it will be hopefully fixed soon.

PR #2108, opened 2020-12-17
chore: setup running tests on iOS

web-padawan:
Run iOS tests without having to use SauceLabs, which should be faster because of less expensive network roundtrips.

See the [instructions](https://github.com/webdriverio/appium-boilerplate/blob/master/docs/APPIUM.md) on ow to install the Appium server and run tests on iOS Simulator locally.


web-padawan:
For some reason appium does not seem to work in GitHub Actions, although locally it works. Will try in a different repo.

PR #2114, opened 2021-01-07
feat: support generics for item type [skip ci]

### LINKED ISSUE #2104: TypeScript generics to allow declaring GridItem type
### Description
Currently the TS type declarations use `type GridItem = unknown;` for the item type in all types/interfaces. It would be great if the types would use generics where the item type could default to GridItem (unknown) but it would be able to be overridden in app code when declaring data providers or renderers for example, so that you don't need to explicitly cast the `unknown` to something else to be able to use item references with TS.

I think all types that use GridItem (directly or nested) should have generics for setting item type.

e.g.
```ts
export type GridDataProviderCallback<T = GridItem> = (
  items: Array<T>,
  size?: number
) => void;

export type GridDataProviderParams<T = GridItem> = {
  page: number;
  pageSize: number;
  filters: Array<GridFilter>;
  sortOrders: Array<GridSorter>;
  parentItem?: T;
};

export type GridDataProvider<T = GridItem> = (
  params: GridDataProviderParams<T>,
  callback: GridDataProviderCallback<T>
) => void;
```

### Expected code for accessing props on custom item type
```ts
const dataProvider: GridDataProvider<MyItem> = (params, callback) => {
  const foo = params.parentItem.myItemProp;
  // ...
};
```

### Actually needed code for accessing props on custom item type
```ts
const dataProvider: GridDataProvider = (params, callback) => {
  const foo = (params.parentItem as MyItem).myItemProp;
  // ...
};
```

web-padawan:
Fixes #2104

This is a breaking change which consists of the following:

1. Removes `GridItem` which is just an alias for `unknown` and therefore is pretty much useless;
2. Change certain types to use generics using the type cast like this:

```ts
const grid = this.shadowRoot.querySelector('vaadin-grid') as GridElement<User>
````

I had to copy-paste and override the related properties on the `GridElement` class instead of modifying relevant mixins. 
This is needed because of the TypeScript limitation, see https://github.com/vaadin/vaadin-combo-box/issues/985#issuecomment-749380363 and https://github.com/microsoft/TypeScript/issues/24122.

### DIFF
--- a/src/vaadin-grid-data-provider-mixin.d.ts
+++ b/src/vaadin-grid-data-provider-mixin.d.ts
@@ -1,17 +1,17 @@
-import { GridDataProvider, GridItem } from './interfaces';
+import { GridDataProvider } from './interfaces';
 
 declare class ItemCache {
   grid: HTMLElement;
   parentCache: ItemCache | undefined;
-  parentItem: GridItem | undefined;
+  parentItem: unknown | undefined;
   itemCaches: object | null;
   items: object | null;
   effectiveSize: number;
   size: number;
   pendingRequests: object | null;
-  constructor(grid: HTMLElement, parentCache: ItemCache | undefined, parentItem: GridItem | undefined);
+  constructor(grid: HTMLElement, parentCache: ItemCache | undefined, parentItem: unknown | undefined);
   isLoading(): boolean;
-  getItemForIndex(index: number): GridItem | undefined;
+  getItemForIndex(index: number): unknown | undefined;
   updateSize(): void;
   ensureSubCacheForScaledIndex(scaledIndex: number): void;
   getCacheAndIndex(index: number): { cache: ItemCache; scaledIndex: number };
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

**Comment by web-padawan** [comment]
Looks like using generics in `GridElement` will break `GridProElement` class declaration 😕

File: src/vaadin-grid-data-provider-mixin.d.ts [hunk 1]
```diff
@@ -51,7 +51,7 @@ interface DataProviderMixin {
    *     are requested, total number of items in the requested sublevel.
    *     Optional when tree is not used, required for tree.
    */
-  dataProvider: GridDataProvider | null | undefined;
+  dataProvider: GridDataProvider<any> | null | undefined;
```
vlukashov, line 54: Should not we try to use `unknown` instead of `any` to force the developer to explicitly specify the type of the data in the grid?

web-padawan:
Closing this for now because:

1. This breaks `vaadin-grid-pro` as mentioned above, and I could not figure out how to fix that,
2. Copying the properties / methods from the mixins to the base class is a really ugly solution.

I don't see how this can be done in a proper way until https://github.com/microsoft/TypeScript/issues/24122 is fixed.

PR #2119, opened 2021-01-13
fix: make explicit data request when clearing cache with no initial pool

yuriy-fix:
Fixes https://github.com/vaadin/vaadin-flow-components/issues/427

### DIFF
--- a/test/data-provider.test.js
+++ b/test/data-provider.test.js
@@ -565,7 +565,7 @@ describe('data provider', () => {
 });
 
 describe('attached', () => {
-  it('should have rows when attached and shown after cache is cleared on hidden grid', (done) => {
+  it('should have rows when attached and shown after cache is cleared on hidden grid', async () => {
     const grid = document.createElement('vaadin-grid');
     const col = document.createElement('vaadin-grid-column');
     col.setAttribute('path', 'item');
@@ -579,15 +579,14 @@ describe('attached', () => {
     grid.style.display = 'none';
     document.body.appendChild(grid);
 
-    setTimeout(() => {
-      grid.clearCache();
-      grid.removeAttribute('style');
-      expect(getCellContent(getFirstVisibleItem(grid)).textContent).to.equal('A');
+    await aTimeout(0);
 
-      // Grid should be removed after test as was attached to body.
-      document.body.removeChild(grid);
-      done();
-    });
+    grid.clearCache();
+    grid.removeAttribute('style');
+    expect(getCellContent(getFirstVisibleItem(grid)).textContent).to.equal('A');
+
+    // Grid should be removed after test as was attached to body.
+    document.body.removeChild(grid);
   });
 });
 
--- a/test/data-provider.test.js
+++ b/test/data-provider.test.js
@@ -8,6 +8,7 @@ import {
   getCellContent,
   getContainerCell,
   getFirstCell,
+  getFirstVisibleItem,
   getRows,
   getRowCells,
   infiniteDataProvider,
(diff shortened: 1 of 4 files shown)


test/sorting.test.js
```diff
@@ -16,6 +18,37 @@ import '../vaadin-grid-sorter.js';
...
+      const col = window.document.createElement('vaadin-grid-column');
+      col.setAttribute('path', 'item');
+      grid.appendChild(col);
+
+      grid.size = 1;
+      grid.dataProvider = function (params, callback) {
+        callback([{ item: 'A' }]);
+      };
+
+      grid.style.maxHeight = '0';
```
tomivirkki, line 33: Is this condition required? If I remove it, the test still passes with the fix and fails without the fix.

test/sorting.test.js
```diff
@@ -16,6 +18,37 @@ import '../vaadin-grid-sorter.js';
...
+      grid.style.display = 'none';
+      window.document.body.appendChild(grid);
+
+      setTimeout(() => {
+        grid.querySelector('vaadin-grid-column').headerRenderer = function (root) {
+          root.innerHTML = '<vaadin-grid-sorter path="item" direction="asc">Item</vaadin-grid-sorter>';
+        };
```
tomivirkki, line 40: If I remove this renderer, the test starts failing. This might indicate that the issue is not specific to sorters and should be fixed somewhere generally.

tulioag (approved):

tomivirkki:
Did some debugging with the original issue and this _might_ be fixable in the data-provider-mixin:

Screenshot 2021-01-15 at 18 25 47

Seems that the `<tbody>` is empty, so initial pool has never been created. Added a debugger to increasePoolIfNeeded...

Screenshot 2021-01-15 at 18 26 20

It was never called...

Screenshot 2021-01-15 at 18 27 01

...so added another one before the guard statement. It got caught a couple of times before the panel was opened, but also once after it got opened. All the other conditions were fine but `canPopulate()` was returning `false`

Screenshot 2021-01-15 at 18 28 06

The `canPopulate()` function checks for `_hasData` and `_columnTree` and out of those two, it was the `_hasData` that was false...

Screenshot 2021-01-15 at 18 28 32

Seems that `_hasData` gets set `true` when a data provider responds...

Screenshot 2021-01-15 at 18 29 14

...so it must have been `true` at some point.

Screenshot 2021-01-15 at 18 30 02

The only place that sets it false is `clearCache()`, and sort-mixin indeed calls `clearCache()` (once the sorters are set).

Screenshot 2021-01-15 at 18 30 34

The end of `clearCache()` makes an explicit data request, but _only_ in the case the grid doesn't have physical items/rows/effectiveSize beforehand. If it does, then it relies on the grid rows to ask for more data and the explicit data request is skipped.

But the condition here seems to be wrong since even though the grid has `_effecticeSize`, there are 0 physical rows rendered at this point and thus the new data request never gets made (and `_hasData` will remain `false`).

So one fix might be to change the condition to also make the explicit data request in case the initial pool hasn't been created:

Screenshot 2021-01-15 at 18 31 55

The PR's original revision works because it explicitly calls `createPool` right before the sort-mixin's `clearCache` manages to change `_hasData` to `false`. It's probably a plausible fix in case sorters are used but an explicit call to `clearCache()` of a hidden grid might still suffer from the same issue. That's probably a good scenario for a test case for the fix.

test/sorting.test.js
yuriy-fix: Removed.

test/sorting.test.js
yuriy-fix: Updated.

test/data-provider.test.js
```diff
@@ -563,6 +564,33 @@ describe('data provider', () => {
   });
 });
 
+describe('attached', () => {
+  it('should have rows when attached and shown after cache is cleared on hidden grid', (done) => {
+    const grid = window.document.createElement('vaadin-grid');
+    const col = window.document.createElement('vaadin-grid-column');
```
web-padawan, line 570:
```suggestion
    const grid = document.createElement('vaadin-grid');
    const col = document.createElement('vaadin-grid-column');
```

test/data-provider.test.js
```diff
@@ -563,6 +564,33 @@ describe('data provider', () => {
...
+    col.setAttribute('path', 'item');
+    grid.appendChild(col);
+
+    grid.size = 1;
+    grid.dataProvider = function (params, callback) {
+      callback([{ item: 'A' }]);
+    };
+
+    grid.style.display = 'none';
+    window.document.body.appendChild(grid);
```
web-padawan, line 580:
```suggestion
    document.body.appendChild(grid);
```

test/data-provider.test.js
```diff
@@ -563,6 +564,33 @@ describe('data provider', () => {
...
+
+    setTimeout(() => {
+      grid.clearCache();
+      grid.removeAttribute('style');
+      expect(getCellContent(getFirstVisibleItem(grid)).textContent).to.equal('A');
+
+      // Grid should be removed after test as was attached to body.
+      window.document.body.removeChild(grid);
```
web-padawan, line 588:
```suggestion
      document.body.removeChild(grid);
```

test/data-provider.test.js
```diff
@@ -563,6 +564,33 @@ describe('data provider', () => {
...
+
+    grid.size = 1;
+    grid.dataProvider = function (params, callback) {
+      callback([{ item: 'A' }]);
+    };
+
+    grid.style.display = 'none';
+    window.document.body.appendChild(grid);
+
+    setTimeout(() => {
```
web-padawan, line 582: nit: please consider using `await aTimeout(0)` and make test  `async () =>` instead of `(done) =>`

web-padawan (commented):
Just a few comments related to the code style.

test/data-provider.test.js
yuriy-fix: Updated.

Issue #1969, opened 2019-05-15
NPE related to sorting when recreating columns


vesanieminen:
The following error happens if Grid columns have sorting enabled and the columns are recreated:

`Caused by: java.lang.IllegalArgumentException: Received a sorters changed call from the client for a non-existent column
	at com.vaadin.flow.component.grid.Grid.sortersChanged(Grid.java:2791)`

![Screen Shot 2019-05-15 at 10 44 44](https://user-images.githubusercontent.com/108755/57757615-7fd5ce80-76fe-11e9-8107-1d58ab1002d6.png)

Steps to reproduce:

1. git clone git@github.com:vesanieminen/Vaadin-Grid-recreate-columns-bug.git
2. mvn jetty:run
3. navigate to localhost:9997
4. click on the sorting buttons on the grid header for one of the columns
5. Select a different value from the Select component.

Expected behaviour:

- The Grid columns are recreated with no exceptions in the backend

Actual behaviour:

- The Grid columns are recreated with an exception.

Workaround exists here: https://github.com/vesanieminen/Vaadin-Grid-recreate-columns-bug/commit/100d56c352f23674e0fbe562878c4473a1e1d3bc

Full error log for the bug:

`[qtp1571205437-69] ERROR com.vaadin.flow.server.DefaultErrorHandler - 
java.lang.RuntimeException: java.lang.IllegalArgumentException: Received a sorters changed call from the client for a non-existent column
	at com.vaadin.flow.server.communication.rpc.PublishedServerEventHandlerRpcHandler.invokeMethod(PublishedServerEventHandlerRpcHandler.java:174)
	at com.vaadin.flow.server.communication.rpc.PublishedServerEventHandlerRpcHandler.invokeMethod(PublishedServerEventHandlerRpcHandler.java:129)
	at com.vaadin.flow.server.communication.rpc.PublishedServerEventHandlerRpcHandler.handleNode(PublishedServerEventHandlerRpcHandler.java:117)
	at com.vaadin.flow.server.communication.rpc.AbstractRpcInvocationHandler.handle(AbstractRpcInvocationHandler.java:64)
	at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocationData(ServerRpcHandler.java:387)
	at com.vaadin.flow.server.communication.ServerRpcHandler.lambda$handleInvocations$1(ServerRpcHandler.java:368)
	at java.util.ArrayList.forEach(ArrayList.java:1251)
[...]

PR #2122, opened 2021-01-20
fix: update filtering and sorting when grid and columns are attached / detached

yuriy-fix:
Part of fix for https://github.com/vaadin/vaadin-grid/issues/1969


src/vaadin-grid-column.js
```diff
@@ -199,6 +199,28 @@ export const ColumnBaseMixin = (superClass) =>
...
+    }
+
+    /** @private */
+    __getColumnFilters() {
+      return this.__getElementsInCells('vaadin-grid-filter');
+    }
+
+    /** @private */
+    __getElementsInCells(name) {
+      // TODO: Check how grid behaves with several filters and sorters in different column cells  to verify this logic
```
tulioag, line 214:
```suggestion
```

tomivirkki: Was this checked?

src/vaadin-grid-dynamic-columns-mixin.js
```diff
@@ -88,10 +88,14 @@ export const DynamicColumnsMixin = (superClass) =>
...
-          info.removedNodes.filter(this._isColumnElement).length > 0
-        ) {
+        const removedColumns = info.removedNodes.filter(this._isColumnElement);
+        if (info.addedNodes.filter(this._isColumnElement).length > 0 || removedColumns.length > 0) {
+          this.__removeSorters(
+            removedColumns.reduce((sorters, column) => sorters.concat(column.__getColumnSorters()), [])
+          );
+          this.__removeFilters(
+            removedColumns.reduce((filters, column) => filters.concat(column.__getColumnFilters()), [])
+          );
```
tomivirkki, line 98: The removed nodes contain the removed sorters/filters (as child elements). You shouldn't need to add any logic to vaadin-grid-column since you can get the removed sorters here instead.

Querying by tag name is also fragile (I remember a user who defined their own tag names for sorters). Shouldn't you be able to access all removed sorters here with `this._sorters.filter((n) => n.isConnected)`?

src/vaadin-grid.js
```diff
@@ -410,10 +410,18 @@ class GridElement extends ElementMixin(
...
+    this.__isConnected = true;
     super.connectedCallback();
     this.recalculateColumnWidths();
   }
 
+  /** @protected */
+  disconnectedCallback() {
+    this.__isConnected = false;
+    super.disconnectedCallback();
+    this.querySelectorAll('vaadin-grid-sorter').forEach((sorter) => (sorter.__preservedOrder = true));
```
tomivirkki, line 422: Shouldn't this logic be in the sorter's disconnectedCallbacks?

src/vaadin-grid.js
```diff
@@ -410,10 +410,18 @@ class GridElement extends ElementMixin(
 
   /** @protected */
   connectedCallback() {
+    this.__isConnected = true;
```
tomivirkki, line 413: Doesn't `this.isAttached` work?

src/vaadin-grid-sorter.js
```diff
@@ -146,6 +146,12 @@ class GridSorterElement extends ThemableMixin(DirMixin(PolymerElement)) {
       _isConnected: {
         type: Boolean,
         value: false
+      },
+
+      /** @private */
+      __preservedOrder: {
```
tomivirkki, line 152: Can you explain the purpose of this property?

src/vaadin-grid-sort-mixin.js
```diff
@@ -51,31 +51,50 @@ export const SortMixin = (superClass) =>
...
+      this.__applySorters();
+    }
 
+    /** @private */
+    __removeSorters(sorters) {
+      if (sorters.length == 0) {
+        return;
+      }
+
+      sorters.forEach((sorter) => this.__updateSorter(sorter, true));
```
tomivirkki, line 66: The boolean flag for the function makes the logic messy.

src/vaadin-grid-sorter.js
```diff
@@ -170,19 +171,31 @@ class GridSorterElement extends ThemableMixin(DirMixin(PolymerElement)) {
...
     }
   }
 
+  /** @private */
+  __isConnectedChanged(isConnected) {
+    if (isConnected) {
+      if (this._wasDisconnected) {
+        this._wasDisconnected = false;
+      } else {
+        this._pathOrDirectionChanged(this.path, this.direction);
```
yuriy-fix, line 194: Calling another observer from `__isConnectedChanged` doesn't seem right. Probably it should be done within multi-property observer, i.e. `__isConnectedOrPathOrDirectionChanged(isConnected, path, direction)`

src/vaadin-grid-sort-mixin.js
```diff
@@ -51,31 +51,66 @@ export const SortMixin = (superClass) =>
...
+    /** @private */
+    __removeSorters(sorters) {
+      if (sorters.length == 0) {
+        return;
+      }
+
+      sorters.forEach((sorter) => {
+        this._removeArrayItem(this._sorters, sorter);
+        if (this.multiSort) {
+          this.__updateSortOrders();
```
yuriy-fix, line 74: Can we update `sortOrders` once all sorters are removed? Not to call it after each sorter is removed.

src/vaadin-grid-sort-mixin.js
```diff
@@ -51,31 +51,66 @@ export const SortMixin = (superClass) =>
     /** @private */
     _onSorterChanged(e) {
       const sorter = e.target;
+      e.stopPropagation();
+      if (this._isUpdatingSorter) return;
+      try {
+        this._isUpdatingSorter = true;
```
yuriy-fix, line 57: Is it possible to add a check for sorter in `__updateSorter` to avoid using flag over here?

test/filtering.test.js
```diff
@@ -252,6 +262,34 @@ describe('array data provider', () => {
     expect(getBodyCellContent(grid, 0, 0).innerText).to.equal('bar');
   });
 
+  // TODO: find better option to check rather than grid._cache.items...
```
yuriy-fix, line 265: In case grid is attached, we could expose and use `getVisibleItems(grid).length` from `helpers`.

test/filtering.test.js
```diff
@@ -252,6 +262,34 @@ describe('array data provider', () => {
...
+    filterFirst.value = 'bar';
+    flushFilters(grid);
+
+    grid.removeChild(grid.firstElementChild);
+    flushGrid(grid);
+
+    expect(Object.keys(grid._cache.items).length).to.equal(3);
+  });
+
+  // TODO: find better option to check rather than grid._cache.items...
```
yuriy-fix, line 276: In case grid is not attached, I guess we can proceed with checking `_cache` or alternatively spy on i.e. updating the item.

src/vaadin-grid-sorter.js
tulioag: `__isConnectedOrPathOrDirectionChanged(isConnected, path, direction)` doesn't seem right to me because `__isConnectedChanged(isConnected)` has extra logic. How about extracting a third method to call from both observers?

src/vaadin-grid-sort-mixin.js
tulioag: Done

src/vaadin-grid-sort-mixin.js
tulioag: Done

test/filtering.test.js
tulioag: Done

test/filtering.test.js
tulioag: Kept code and removed TODO

src/vaadin-grid-sorter.js
tulioag: Extracted method

src/vaadin-grid-filter-mixin.js
```diff
@@ -29,13 +29,38 @@ export const FilterMixin = (superClass) =>
...
     _filterChanged(e) {
-      if (this._filters.indexOf(e.target) === -1) {
-        this._filters.push(e.target);
+      e.stopPropagation();
+
+      this.__updateFilter(e.target);
+      this.__applyFilters();
+    }
+
+    __removeFilters(filters) {
```
sissbruecker, line 38:
```suggestion
    /** @private */
    __removeFilters(filters) {
```

src/vaadin-grid-filter-mixin.js
```diff
@@ -29,13 +29,38 @@ export const FilterMixin = (superClass) =>
...
+      if (filters.length == 0) {
+        return;
       }
 
-      e.stopPropagation();
+      this._filters.forEach((filter) => {
```
sissbruecker, line 43: This looks like a bug: The `filters` param of the method is not used in this loop. Instead it removes every single filter in `this._filters`. It looks like you want to loop on `filters` instead of `this._filters`.

Also wondering if it would not be possible to use `Array.filter` here? Basically:
```
this._filters = this._filters.filter(filter => filters.indexOf(filter) >= 0);
```

Might also be more clear to rename the param to `filtersToRemove`.

src/vaadin-grid-filter-mixin.js
```diff
@@ -29,13 +29,38 @@ export const FilterMixin = (superClass) =>
...
+        const filterIndex = this._filters.indexOf(filter);
+        if (filterIndex !== -1) {
+          this._filters.splice(filterIndex, 1);
+        }
+      });
+      this.__applyFilters();
+    }
+
+    /** @private */
+    __updateFilter(filter) {
```
sissbruecker, line 53: `__updateFilter` sounds confusing, since it only adds. What is the intent here?

src/vaadin-grid-sort-mixin.js
```diff
@@ -51,31 +51,63 @@ export const SortMixin = (superClass) =>
...
+    }
+
+    /** @private */
+    __removeSorters(sorters) {
+      if (sorters.length == 0) {
+        return;
+      }
+
+      sorters.forEach((sorter) => {
+        this._removeArrayItem(this._sorters, sorter);
```
sissbruecker, line 66: This uses a different method of removing items from an array than what is used above in `__removeFilters`, I would suggest to align them.

src/vaadin-grid-sorter.js
```diff
@@ -170,19 +171,34 @@ class GridSorterElement extends ThemableMixin(DirMixin(PolymerElement)) {
...
-  _pathOrDirectionChanged(path, direction, isConnected) {
-    if (path === undefined || direction === undefined || isConnected === undefined) {
-      return;
-    }
+  _pathOrDirectionChanged() {
+    this.__dispatchSorterChangedEvenIfPossible();
+  }
 
+  /** @private */
+  __isConnectedChanged(isConnected) {
```
sissbruecker, line 183: Does this receive the old value as well? If so the `_wasDisconnected` construct might not be necessary, and you can just check if old and new value differ.

test/filtering.test.js
```diff
@@ -107,6 +115,12 @@ describe('filtering', () => {
     grid._filters.forEach((filter) => expect(filter.$.filter.clientHeight).to.be.greaterThan(0));
   });
 
+  it('should not keep references to filters when column is removed', () => {
```
sissbruecker, line 118: Considering the probably bug above, we might want to add a test case where we check that it keeps the filters for columns that were not removed.

src/vaadin-grid-sorter.js
tulioag: Good point. Removed `_wasDisconnected`

src/vaadin-grid-filter-mixin.js
tulioag: Done

src/vaadin-grid-filter-mixin.js
tulioag: Renamed to `__addFilter`

src/vaadin-grid-sort-mixin.js
tulioag: Done

src/vaadin-grid-dynamic-columns-mixin.js
tulioag: I refactored not to query tag names, but `this._sorters.filter((n) => !n.isConnected)` was not enough.

sissbruecker (approved):

PR #2131, opened 2021-02-16
Flush debounceIncreasePool on loadPage if scrolling

### LINKED ISSUE #2107: TreeGrid regression since 14.4.3
### Description
TreeGrid since 14.4.3 displays blank areas when an initial scroll happens.

### Expected outcome
Scroll should work. Rows are displayed .
grafik

### Actual outcome
grafik

### Steps to reproduce
```
public class GridScrollUi extends HorizontalLayout implements AfterNavigationObserver {

    private final List<Double> values;

    public GridScrollUi() {
        super();
        setDefaultVerticalComponentAlignment(Alignment.STRETCH);
        values = DoubleStream.generate(Math::random)
                .filter(item -> item != Double.MIN_VALUE)
                .limit(10000)
                .boxed()
                .collect(Collectors.toList());
    }

    @Override
    public void afterNavigation(AfterNavigationEvent afterNavigationEvent) {
        VerticalLayout layout = new VerticalLayout();
        layout.setMaxWidth("300px");
        layout.setMaxHeight("300px");

        List<Double> roots = new ArrayList<>();
        roots.add(Double.MIN_VALUE);

        TreeGrid<Double> treeGrid = new TreeGrid<>();
        treeGrid.addHierarchyColumn(item -> item);
        treeGrid.setItems(roots, value -> {
            if (value == null) {
                return roots;
            } else if (value == Double.MIN_VALUE) {
                return values;
            } else {
                return new ArrayList<>();
            }
        });
        layout.add(treeGrid);
        add(layout);

        // Initial scroll to bottom
        treeGrid.expand(Double.MIN_VALUE);
        int pos = values.size() - 1;
        treeGrid.select(values.get(pos));

        boolean isAtLeast14_4_3 = true;
        if (isAtLeast14_4_3) {
            treeGrid.scrollToIndex(pos);
        } else {
            // Vaadin < 14.4.3 needs this workaround and it more or less worked
            // see https://vaadin.com/forum/thread/18447912/possible-regression-in-new-grid-scroll-implementation
            // and https://github.com/vaadin/vaadin-grid/issues/1808
            treeGrid.getElement().executeJs("setTimeout(function() { $0.scrollToIndex($1) })", treeGrid.getElement(), pos);
            // Since there are many items, a delay is needed
            treeGrid.getElement().executeJs("setTimeout(function() { $0.scrollToIndex($1) }, 1500)", treeGrid.getElement(), pos);
        }
    }

}
```

### Browsers Affected

- [ X  ] Chrome
- [ X ] Firefox
- [ X ] Edge NEW

- [ ? ] Safari
- [ ? ] IE 11
- [ ? ] iOS Safari
- [ ? ] Android Chrome

tulioag:
Fixes #2107

### DIFF
--- a/src/vaadin-grid-data-provider-mixin.html
+++ b/src/vaadin-grid-data-provider-mixin.html
@@ -345,7 +345,7 @@ This program is available under Apache License Version 2.0, available at https:/
           filters: this._mapFilters(),
           parentItem: cache.parentItem
         };
-
+        this._debounceIncreasePool && this._debounceIncreasePool.flush();
         this.dataProvider(params, (items, size) => {
           if (size !== undefined) {
             cache.size = size;
--- a/test/data-provider.html
+++ b/test/data-provider.html
@@ -366,7 +366,7 @@
             // Effective size should change in between the data requests
             expect(renderSpy.called).to.be.true;
             expect(increasePoolSpy.callCount).to.above(1);
-            expect(updateItemSpy.callCount).to.be.below(90);
+            expect(updateItemSpy.callCount).to.be.below(180);
           });
 
           it('should keep item expanded on itemIdPath change', () => {
--- a/test/scroll-to-index.html
+++ b/test/scroll-to-index.html
@@ -205,7 +205,6 @@
           };
         });
 
-
       });
 
       describe('Added item', () => {
--- a/test/scroll-to-index.html
+++ b/test/scroll-to-index.html
@@ -39,6 +39,13 @@
     </template>
   </test-fixture>
 
+  <test-fixture id="treeGrid">
+    <template>
+      <vaadin-grid style="width: 200px; height: 500px;">
+        <vaadin-grid-tree-column path="name" header="foo" item-has-children-path="hasChildren"></vaadin-grid-tree-column>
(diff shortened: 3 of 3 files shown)


src/vaadin-grid-data-provider-mixin.html
```diff
@@ -345,7 +345,9 @@
           filters: this._mapFilters(),
           parentItem: cache.parentItem
         };
-
+        if (this.__scrollingToPendingIndex && this._debounceIncreasePool) {
```
tomivirkki, line 348: This new `__scrollingToPendingIndex` property feels a bit hacky? Is there some other justification for adding it other than making that one data-provider test pass / should the increase pool debouncer not be flushed while the grid is scrolling to pending index?

I took a look at the failing test case and it seems to pass because of timing. If I wrap the last expect inside of a timeout of 1 second, it also fails on master due to a too large call count. I don't necessarily see it as a problem to increase the call count expectation to make the test pass with the new change in timing.

tulioag: Done

tomivirkki (approved):

PR #2138, opened 2021-02-26
Update web-test-runner, refactor tests

web-padawan:


tomivirkki (approved):

test/array-data-provider.test.js
```diff
@@ -224,45 +224,28 @@ describe('invalid paths', () => {
...
-      listenOnce(filter, 'filter-changed', () => {
-        setTimeout(() => {
-          expect(console.warn.called).to.be.false;
-          done();
-        });
-      });
-
+    it('should not warn about undefined values with defined parent property', () => {
       filter.path = 'name.foo';
+      filter._debouncerFilterChanged.flush();
```
tomivirkki, line 241: nit: is it intentional to flush a specific debouncer instead of running `flushGrid(grid)`?

web-padawan: Yes, this debouncer is not present in `flushGrid` method. 
Alternatively we could use fake timers, but that would mean having to use magic numbers in tests.

Issue #1960, opened 2019-06-07
Expansion state is lost when reattaching TreeGrid


stefanuebe:
The TreeGrid loses it's nodes' expansion states (i.e. everything up to the root is collapsed), when being reattached. 

Use case where this has an impact on the user experience:
A view with a tree, that is annotated as UIScoped in a CDI context (e.g. Spring). The user works in the tree, expands several nodes and switches the page (for whatever reason). When coming back, the previously instantiated view instance is reused and attached. Other states (selection for instance) are restored, the expansion states are not. The user has now to reopen all nodes as needed.

Request is, that the tree should keep it's expansion states to be reusable when reattached.

web-padawan:
Should be fixed by vaadin/vaadin-flow-components#363. See also #1820 where the fix was discussed.

PR #2179, opened 2021-05-19
fix: disable grid section focus target in interaction mode (#1960) (CP 19.0)

sissbruecker:
Fixes: vaadin/vaadin-select#265
Warranty: Fixes moving focus between inputs in adjacent cells when in interaction mode

### DIFF
--- a/test/keyboard-navigation.test.js
+++ b/test/keyboard-navigation.test.js
@@ -25,6 +25,7 @@ import '../vaadin-grid-column-group.js';
 
 // Detect if tests are running in Sauce Labs, see `web-test-runner.config.js`
 const isSauceLabsLauncher = window.browserLauncher === 'saucelabs';
+const itNoSauce = isSauceLabsLauncher ? it.skip : it;
 
 window.top.focus && window.top.focus();
 window.focus();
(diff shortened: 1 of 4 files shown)


test/keyboard-navigation.test.js
```diff
@@ -1556,14 +1613,51 @@ describe('keyboard navigation', () => {
       spy.restore();
     });
 
-    it('should focus the next input element when tabbing in interaction mode', () => {
-      right(); // focus the cell with input.
+    // Skip: sendKeys is not supported in WTR when using web-driver
+    it.skip('should focus the next input element when tabbing in interaction mode', async () => {
```
sissbruecker, line 1617: Decided to disable the (3) relevant tests for this as the sendKeys command is not supported in WTR when using web-driver:
```
Error: Error while executing command send-keys with payload {"press":"Tab"}: Sending keys is not supported for browser type webdriver.
```
I can not think of a reasonable way to test the behaviour in a different manner, as it's supposed to test that the implicit tab order works correctly when using the tab key.

I also decided to still install the plugin and keep the code here uncommented / use `skip`, instead of just uncommenting / removing the code. Not sure what's best practice, can also comment out the code instead.

sissbruecker:
@tulioag When reviewing please note that I decided to disable 3 tests here, see comment above.

tulioag (approved):

tulioag (approved):

PR #2184, opened 2021-06-15
feat: deprecate heightByRows in favor of allRowsVisible

fhdhsni:
related to: vaadin/flow-components#1009

- [X] Feature

## Checklist

- [X] I have added tests to ensure my change is effective and works as intended.
- [X] New and existing tests are passing locally with my change.
- [X] I have performed self-review and corrected misspellings.

### DIFF
--- a/demo/grid-data-demos.html
+++ b/demo/grid-data-demos.html
@@ -54,11 +54,11 @@
 
     <h3>Dynamic Height</h3>
     <p>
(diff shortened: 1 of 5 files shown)


sissbruecker:
Vaadin Grid 5.7 is also used in [platform 14.6](https://github.com/vaadin/platform/blob/14.6/versions.json#L192). If we want this change to only land in 14.7, we need to:
- create Vaadin Grid version 5.8
- update the platform to use 5.8 for 14.7

yuriy-fix (approved):

PR #2194, opened 2021-08-11
fix: column widths are properly resized from the first time it is exanded

sosa-vaadin:
## Description

There's a timing issue between the rendering of columns and the call to `recalculateColumnWidths()` when a tree toggle is clicked. This caused the method to be called before the tree columns were rendered and so would calculate their widths improperly (since no content is available, it will use much less space than the content actually needed).

The timing issue is fixed by adding a new method to check if iron-list is still creating rows and using it to determine when the `recalculateColumnWidths()` should be called.

In addition, the `level` CSS style must be set synchronously since it was also not being applied in time. When it did apply, the column width was already recalculated and then it would clip the content by the level-spacer's width.

Fixes https://github.com/vaadin/flow-components/issues/1855

## Type of change

- [x] Bugfix

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

### DIFF
--- a/src/vaadin-grid-tree-toggle.html
+++ b/src/vaadin-grid-tree-toggle.html
@@ -195,6 +195,13 @@ This program is available under Apache License Version 2.0, available at https:/
       _levelChanged(level) {
         const value = Number(level).toString();
         this.updateStyles({'---level': value});
+        // Async is to make DOM updates applied before evaluating the style
+        // update, required for polyfilled RTL support in MSIE and Edge.
+        this._debouncerUpdateLevel = Polymer.Debouncer.debounce(
+          this._debouncerUpdateLevel,
+          Polymer.Async.microTask,
+          () => this.updateStyles({'---level': value})
+        );
       }
     }
 
--- a/src/vaadin-grid-tree-toggle.html
+++ b/src/vaadin-grid-tree-toggle.html
@@ -194,14 +194,7 @@ This program is available under Apache License Version 2.0, available at https:/
       /** @private */
       _levelChanged(level) {
         const value = Number(level).toString();
-        this.style['---level'] = value;
-        // Async is to make DOM updates applied before evaluating the style
-        // update, required for polyfilled RTL support in MSIE and Edge.
-        this._debouncerUpdateLevel = Polymer.Debouncer.debounce(
-          this._debouncerUpdateLevel,
-          Polymer.Async.microTask,
-          () => this.updateStyles({'---level': value})
-        );
+        this.updateStyles({'---level': value});
       }
     }
 
--- a/src/vaadin-grid.html
+++ b/src/vaadin-grid.html
@@ -457,10 +457,12 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       /** @protected */
+      // This function is more like "tryToResizeWidth" instead
       __itemsReceived() {
         if (this._recalculateColumnWidthOnceLoadingFinished
-          && !this._cache.isLoading()
-          && this.__hasRowsWithClientHeight()) {
+        && !this._cache.isLoading()
+        && this.__hasRowsWithClientHeight()
+            && (!this._debounceIncreasePool || !this._debounceIncreasePool.isActive())) {
           this._recalculateColumnWidthOnceLoadingFinished = false;
           this.recalculateColumnWidths();
         }
--- a/src/vaadin-grid.html
+++ b/src/vaadin-grid.html
@@ -457,7 +457,6 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       /** @protected */
-      // This function is more like "tryToResizeWidth" instead
       __flushPendingRecalculateColumnWidths() {
         if (this._recalculateColumnWidthOnceLoadingFinished
         && !this._cache.isLoading()
--- a/src/vaadin-grid.html
+++ b/src/vaadin-grid.html
@@ -458,7 +458,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
       /** @protected */
       // This function is more like "tryToResizeWidth" instead
-      __itemsReceived() {
+      __flushPendingRecalculateColumnWidths() {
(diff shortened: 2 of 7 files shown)


src/vaadin-grid-tree-toggle.html
```diff
@@ -194,14 +194,7 @@
...
         const value = Number(level).toString();
-        this.style['---level'] = value;
-        // Async is to make DOM updates applied before evaluating the style
-        // update, required for polyfilled RTL support in MSIE and Edge.
-        this._debouncerUpdateLevel = Polymer.Debouncer.debounce(
-          this._debouncerUpdateLevel,
-          Polymer.Async.microTask,
-          () => this.updateStyles({'---level': value})
-        );
+        this.updateStyles({'---level': value});
```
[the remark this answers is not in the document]
sosa-vaadin: I ran all tests in Internet Explorer in a Windows 10 machine. All 994 tests passed in 254 seconds.

src/vaadin-grid.html
```diff
@@ -457,10 +457,12 @@
       }
 
       /** @protected */
-      __itemsReceived() {
+      // This function is more like "tryToResizeWidth" instead
```
tomivirkki, line 460: This change should be reverted

test/tree-toggle.html
```diff
@@ -129,7 +129,6 @@
           let prevWidth = 0;
           for (let i = 1; i < 3; i++) {
             toggle.level = i;
-            toggle._debouncerUpdateLevel.flush();
```
tomivirkki, line 132: This change should be reverted

src/vaadin-grid.html
```diff
@@ -457,10 +457,11 @@
...
 
       /** @protected */
-      __itemsReceived() {
+      __flushPendingRecalculateColumnWidths() {
         if (this._recalculateColumnWidthOnceLoadingFinished
-          && !this._cache.isLoading()
-          && this.__hasRowsWithClientHeight()) {
+        && !this._cache.isLoading()
+        && this.__hasRowsWithClientHeight()
+            && (!this._debounceIncreasePool || !this._debounceIncreasePool.isActive())) {
```
tomivirkki, line 464: Adding a debugger to `_recalculateColumnWidths` shows that it's invoked too often during the test. The reason is that the check on this row is missing from `recalculateColumnWidths()` which gets directly invoked in the test. I'd suggest creating a separate `__isCreatingRows()` function if you're going to have the check in 2 places.

Drafted the following proposal, ptal:
```js
__isCreatingRows() {
        // iron-list's _increasePoolIfNeeded function may have scheduled a debouncer that will re-invoke _increasePoolIfNeeded (in vaadin-grid-scroller)
        const schedulingDebounceIncreasePool = this._debouncers && this._debouncers._increasePoolIfNeeded && this._debouncers._increasePoolIfNeeded.isActive();
        // vaadin-grid-scroller's overridden _increasePoolIfNeeded may have scheduled a _debounceIncreasePool debouncer
        const debouncingIncreasePoolActive = this._debounceIncreasePool && this._debounceIncreasePool.isActive();

        // If either of the conditions are true, we consider the grid is still in the process of creating new rows
        return schedulingDebounceIncreasePool || debouncingIncreasePoolActive;
}
```

Notice how in the proposal, the column width calculation is blocked also when iron-list may have scheduled a debouncer to re-invoke the function. This means that `_increasePoolIfNeeded` debounceer may actually be inactive. So you'll need to add another call to `__flushPendingRecalculateColumnWidths` in vaadin-grid-scroller when the function is invoked but `_debounceIncreasePool` may have not gotten activated. I wonder if another test is required?

test/column-auto-width.html
```diff
@@ -111,6 +111,33 @@
...
+        grid.clearCache();
+
+        // Invoke recalculateColumnWidths manually
+        //
+        // Before the fix, this would have executed immediately, since all the data items are already loaded,
+        // but since the DOM row elements haven't yet been created, the column widths would not get calculated
+        // from the expected set of samples.
+        // With the fix in place, the recalculateColumnWidths will not execute until the grid
+        // has also finished creating the necessary DOM row elements.
+        // debugger;
```
tomivirkki, line 133: Unintentional comment

src/vaadin-grid-tree-toggle.html
```diff
@@ -194,7 +194,7 @@
       /** @private */
       _levelChanged(level) {
         const value = Number(level).toString();
-        this.style['---level'] = value;
+        this.updateStyles({'---level': value});
```
tomivirkki, line 197: Can this change be tested?

tomivirkki (changes requested):
The PR description is outdated. It describes another approach to fixing the issue but it's no longer used in this PR.

test/column-auto-width.html
sosa-vaadin: I removed the debugger comment.

src/vaadin-grid-tree-toggle.html
sosa-vaadin: I added a test for this change.

tomivirkki: nit: it merely tests that the width gets synchronously changed. The width changing asynchronously isn't a bug per se. The bug, in this case, is that `recalculateColumnWidths()` doesn't wait until a tree-toggle's `level` change gets the styles properly applied but executes too soon.

Ideally, the test should verify that the bug caused by the width changing asynchronously, got fixed by the change = a test that adds a tree-toggle to a cell, applies a level-value for it, and invokes `recalculateColumnWidths()`.

Not a blocker for this PR though.

tomivirkki (approved):

PR #2200, opened 2021-11-26
fix: correctly show rows after showing hidden column and clearing grid filter

hdamr:
## Description

When grid is used with an in-memory data provider with more than 50 items, after applying a filter and hiding a column and respectively making the column visible and filter cleared (in this particular order),  grid breaks and it won't show items correctly unless page is refreshed. 

The current fix simply resolves by checking for cell parent not being null in `__setColumnTemplateOrRenderer` method, and by doing so the issue is resolved. I have added a relevant test case which tests this particular behaviour.

This PR is rebased on top of master (which includes all changes made into running tests)

Fixes https://github.com/vaadin/flow-components/issues/2330

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

### DIFF
--- a/src/vaadin-grid-column.html
+++ b/src/vaadin-grid-column.html
@@ -322,6 +322,9 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       cells.forEach(cell => {
+        if (!cell.parentElement) {
+          return;
+        }
         const model = this._grid.__getRowModel(cell.parentElement);
 
         if (renderer) {
--- a/src/vaadin-grid-column.html
+++ b/src/vaadin-grid-column.html
@@ -322,10 +322,15 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       cells.forEach(cell => {
-        if (!cell.parentElement) {
+        const parent = cell.parentElement;
+
+        // When a column is made hidden and shown again, in some instances it breaks rendering of rows for grid
+        // this happens when parent element of cell is null, which might not be set correctly during rendering
+        // the newly shown column, this check simply avoid that case
+        if (!parent) {
           return;
         }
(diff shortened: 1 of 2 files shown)


sissbruecker (approved):

src/vaadin-grid-column.html
```diff
@@ -322,6 +322,9 @@
       }
 
       cells.forEach(cell => {
+        if (!cell.parentElement) {
```
web-padawan, line 325: Let's update the code to only check for `cell.parentElement` once:

```js
const parent = cell.parentElement;
if (!parent) {
  return;
}
const model = this._grid.__getRowModel(parent);
```

src/vaadin-grid-column.html
```diff
@@ -322,6 +322,9 @@
...
+          return;
```
web-padawan, line 326: Let's add a comment to briefly mention when this could happen and cause problems.

web-padawan (changes requested):

src/vaadin-grid-column.html
hdamr: done.

src/vaadin-grid-column.html
hdamr: done.

web-padawan (approved):

PR #2201, opened 2021-11-29
CI test with browsers separated into different steps

hdamr:
THIS IS FOR TEST!


hdamr:
This was investigated and resolved.

PR #2205, opened 2022-01-14
chore: add lockfiles to ensure consistent builds in CI

web-padawan:
## Description

Connected to https://github.com/vaadin/components-team-tasks/issues/602
Depends on https://github.com/vaadin/magi-cli/pull/158

## Type of change

- Internal change


tomivirkki (approved):

src/vaadin-grid-column-reordering-mixin.html
```diff
@@ -215,15 +215,24 @@
...
 
-        // Workaround a FF58 bug
+      this._reorderGhost.hidden = false;
+
+      // Workaround FF58/ShadyDOM bugs
+      while (!cell._column && cell.parentElement) {
         if (cell.localName === 'vaadin-grid-cell-content') {
           cell = cell.assignedSlot.parentNode;
+        } else {
+          cell = cell.parentElement;
```
tomivirkki, line 235: I eventually discovered a way to debug the failing tests locally. Turns out the failures were caused by issues in the `elementFromPoint` polyfill and also due to grid assuming `elementFromPoint` not being polyfilled at all (which was the case with older ShadyDOM). Phew...

PR #2214, opened 2022-04-21
refactor: move grid initialization to beforeEach

sissbruecker:
Moves the test setup to beforeEach, in order to avoid possible side effects when modifying or moving the grid in individual test cases.

### DIFF
--- a/test/column-groups.html
+++ b/test/column-groups.html
@@ -253,6 +253,7 @@
   <script>
     describe('column groups', () => {
       let grid, scroller, header, body, footer, headerRows, footerRows, bodyRows;
+      const isFirefox = window.navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
 
       function getFooterCellContent(row, col) {
         return getCellContent(getFooterCell(row, col)).textContent.trim();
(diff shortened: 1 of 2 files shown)


test/keyboard-navigation.html
```diff
@@ -812,7 +783,7 @@
 
             expect(getFocusedCellIndex()).to.equal(1);
             const columnFocusedCell = grid.shadowRoot.activeElement._column;
-            expect(columnFocusedCell.hidden).to.be.false;
+            expect(columnFocusedCell.hidden).not.to.be.true;
```
sissbruecker, line 786: `false` is not the default value for `hidden`, which seems to be undefined. To be safe I switched the assertion the check against `true`.

test/column-groups.html
```diff
@@ -530,7 +531,7 @@
           ].forEach(a => expect(getFooterCell(a[0], a[1]).hasAttribute('frozen')).to.be.true);
         });
 
-        it('should have last row fully visible when scrolled to end', done => {
+        (isFirefox ? it.skip : it)('should have last row fully visible when scrolled to end', done => {
```
sissbruecker, line 534: This is actually broken in Firefox: https://github.com/vaadin/web-components/issues/3717

test/resizing.html
```diff
@@ -108,14 +108,17 @@
           });
         });
 
-        it('should update footer height', () => {
+        it('should update footer height', (done) => {
```
sissbruecker, line 111: Fixes a flaky test in Firefox

PR #2216, opened 2022-05-10
fix: prevent overriding column hidden state when attaching column group (CP: 14.8)

sissbruecker:
Fixes the column group to not synchronize its hidden state to child columns when it is attached, and visible.

Fixes vaadin/flow-components#2959
Cherry-pick of: https://github.com/vaadin/web-components/pull/3805

### DIFF
--- a/src/vaadin-grid-column-group.html
+++ b/src/vaadin-grid-column-group.html
@@ -85,11 +85,8 @@ This program is available under Apache License Version 2.0, available at https:/
 
         static get observers() {
           return [
-            '_updateVisibleChildColumns(_childColumns)',
-            '_childColumnsChanged(_childColumns)',
             '_groupFrozenChanged(frozen, _rootColumns)',
-            '_groupHiddenChanged(hidden, _rootColumns)',
-            '_visibleChildColumnsChanged(_visibleChildColumns)',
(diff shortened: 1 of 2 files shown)


src/vaadin-grid-column.html
```diff
@@ -313,10 +313,6 @@
 
     /** @private */
     __setColumnTemplateOrRenderer(template, renderer, cells) {
-      // no renderer or template needed in a hidden column
-      if (this.hidden) {
-        return;
-      }
```
sissbruecker, line 319: The change seems to affect when the hidden state changes, so this condition was causing header templates not to render. The problem with this condition is that the observer does not listen to hidden changes, so even if the column / group becomes visible this is never called again. Adding hidden to the observer also didn't improve things, as there are some other checks that prevent setting the template.

The condition here was introduced as part of this bugfix: https://github.com/vaadin/vaadin-grid/pull/2146. However it seems the fix was not covering all cases, so a another fix was added later: https://github.com/vaadin/vaadin-grid/pull/2200. It seems that with the second fix the first one is not needed anymore, so it looks like we can remove it. The other concern is doing the extra work of rendering the cells even though they are hidden, though that was already the case before the first fix.

tomivirkki (approved):

